PR #88, opened 2023-06-28
Drawio in vpc model

haim-kermany:
integrated drawio into vpc model
updating tests in ibmvpc


pkg/vpcmodel/abstractVPC.go
```diff
@@ -21,6 +21,9 @@ type NamedResourceIntf interface {
 	UID() string
 	Name() string
 }
+type ZonalNamedResourceIntf interface {
+	ZoneName() string
+}
```
adisos, line 26: might be simpler to add `ZoneName()` to the `NamedResourceIntf` ?

adisos (commented):
looks good, just added one comment

pkg/vpcmodel/abstractVPC.go
haim-kermany: not all NamedResourceIntf  has zone, only the one that inherit zonalNamedResource

haim-kermany: fixed, removed ZonalNamedResourceIntf  and ZonalNamedResource

adisos (approved):

PR #99, opened 2023-07-09
Genetare architecure drawio the second

haim-kermany:


pkg/ibmvpc/analysis_output_test.go
```diff
@@ -234,7 +235,12 @@ func TestAllWithComparison(t *testing.T) {
 	// tests is the list of tests to run
 	for testIdx := range tests {
 		tt := tests[testIdx]
-		tt.mode = outputComparison
+		// todo - remove this if when drawio is stable
+		if tt.format == vpcmodel.DRAWIO || tt.format == vpcmodel.ARCHDRAWIO {
+			tt.mode = outputIgnore
+		} else {
+			tt.mode = outputComparison
+		}
```
adisos, line 243: should also apply this to the commented-out function above (`TestAllWithGeneration`). [this function is there to be used when generating expected output files that override the current files.. it should not be generated for this format until comparison is activated as well).

cmd/analyzer/main.go
```diff
@@ -56,14 +58,14 @@ func _main(cmdlineArgs []string) error {
...
 
-	og, err := vpcmodel.NewOutputGenerator(cloudConfig, *inArgs.Grouping, analysisTypeToUseCase(inArgs))
-	if err != nil {
-		return err
-	}
 	outFile := ""
 	if inArgs.OutputFile != nil {
 		outFile = *inArgs.OutputFile
 	}
+	og, err := vpcmodel.NewOutputGenerator(cloudConfig, *inArgs.Grouping, analysisTypeToUseCase(inArgs), outFile != ARCHDRAWIOFormat)
```
adisos, line 65: why comparing outFile and not the outputFormat to `ARCHDRAWIOFormat` ?

haim-kermany: fixed

adisos (approved):

PR #113, opened 2023-07-25
Issue 112 support empty squares

haim-kermany:


pkg/drawio/layout.go
```diff
@@ -171,16 +191,16 @@ func (*layoutS) resolveSquareLocation(tn SquareTreeNodeInterface, addInternalBor
 }
 
 func (ly *layoutS) setSquaresLocations() {
+	ly.resolveSquareLocation(ly.network, networkToSubnetDepth, false)
```
zivnevo, line 194: Why was the order of calling `resolveSquareLocation` changed?
Don't we want to resolve the location of a parent only after resolving the locations of all its child nodes?

haim-kermany: since some squares does not have children, we can not relay on location of all child nodes.
the new approach is more robust in these two phases:
1. add all the border layers: addAllBorderLayers()
2.  extend the square by setting the border layers of the square  to be **ToSubnetDepth long from its icons location (or its default location)

zivnevo: Got it.

zivnevo (approved):

PR #123, opened 2023-08-02
Comparisons between expected to actual should be robust and oblivious…

ShiriMoran:
… to line separators which are editor dependent

solves https://github.com/np-guard/vpc-network-config-analyzer/issues/118


pkg/ibmvpc/analysis_output_test.go
```diff
@@ -354,6 +354,11 @@ func runTestPerUseCase(t *testing.T, tt *vpcGeneralTest, c *vpcmodel.CloudConfig
 	return nil
 }
 
+// comparison should be insensitive to line comparators; cleaning strings from line comparators
+func cleanStr(str string) string {
+	return strings.ReplaceAll(strings.ReplaceAll(str, "/n", ""), "\r", "")
```
adisos, line 359: why `/n` and not `\n` ?

Issue #103, opened 2023-07-12
handle unidirectional connections under grouping


adisos:
Open an issue for handling of unidirectional connections under grouping

_Originally posted by @zivnevo in https://github.com/np-guard/vpc-network-config-analyzer/pull/96#discussion_r1259439551_

zivnevo:
Done

PR #125, opened 2023-08-03
103 unidirectional grouping

ShiriMoran:


ShiriMoran:
solves #103

zivnevo (commented):
Make sure `md` and `json` formats are also handled properly

pkg/vpcmodel/grouping.go
```diff
@@ -129,13 +146,14 @@ func (g *GroupConnLines) groupExternalAddresses() {
 				continue
 			}
 			connString := conns.String()
+			connIsStateful := fmt.Sprintf("%s;;%d", connString, conns.IsStateful)
```
zivnevo, line 149: Variable name is a bit misleading. How about `connWithStatefulness`?

ShiriMoran: Done, also in other places I used this (or a very similar) name

pkg/vpcmodel/grouping.go
```diff
@@ -1,22 +1,34 @@
...
 
 func (g *groupingConnections) getGroupedConnLines(isSrcToDst bool) []*GroupedConnLine {
 	res := []*GroupedConnLine{}
 	for a, aMap := range *g {
-		for conn, b := range aMap {
+		for connIsStateful, b := range aMap {
+			connIsStatefulSlice := strings.Split(connIsStateful, ";;")
+			if len(connIsStatefulSlice) != 2 {
+				err := fmt.Sprintf("something wrong connIsStatefulSlice %+v is not in the right format; ", connIsStateful)
+				panic(err)
```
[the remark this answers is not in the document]
ShiriMoran: Done. Quit a fuss for an error that will never occur. Its in a separate commit, up for Adi to decide whether to merge the commit.

pkg/vpcmodel/textOutput.go
adisos, line 1: Instead of adding `asteriskDetails` from both functions called from lines 14, 16 (with and without grouping) -- add after line 17.

ShiriMoran: Yes, it crossed my mind.
The thing is that in the grouping case `asteriskDetails` is added by `func (g *GroupConnLines) String() string` which is also used by the grouping's tests (`grouping_test.go`).

adisos (approved):

Issue #100, opened 2023-07-10
merging overlapping/touching IP-blocks


adisos:
- [x]               Please open an issue on merging overlapping/touching IP-blocks. It can improve results here. E.g., replacing `161.26.0.0/16,161.27.0.0/16` with `161.26.0.0/15`
_Originally posted by @zivnevo in https://github.com/np-guard/vpc-network-config-analyzer/pull/95#discussion_r1254196729_

- [x] grouping of ip-blocks should be always applied

ShiriMoran:
2nd item solved by #119

ShiriMoran:
3rd item solved by #132

PR #132, opened 2023-08-08
100 cidr block merge

ShiriMoran:
solves https://github.com/np-guard/vpc-network-config-analyzer/issues/100


ShiriMoran:
solves #100

pkg/common/CanonicalIntervalSet.go
```diff
@@ -84,6 +84,13 @@ func (i *Interval) intersection(other Interval) []Interval {
 	return []Interval{{Start: maxStart, End: minEnd}}
 }
 
+// ToIPRange returns a string of the ip range of a single interval
+func (i *Interval) toIPRange() string {
+	startIP := InttoIP4(i.Start)
+	endIP := InttoIP4(i.End)
+	return rangeIPstr(startIP, endIP)
+}
+
```
adisos, line 93: convert to a function `intervalToIPRange(i *Interval)` within `ipBlock.go`, so that it is an internal function within the `ipblock` pkg rather than a method of `Interval`.

pkg/common/ipBlock.go
```diff
@@ -29,13 +29,15 @@ type IPBlock struct {
 
 // ToIPRanges returns a string of the ip ranges in the current IPBlock object
 func (b *IPBlock) ToIPRanges() string {
+	return strings.Join(b.ToIPRangesList(), ",")
+}
+
+func (b *IPBlock) ToIPRangesList() []string {
```
adisos, line 35:
```suggestion
// ToIPRangesList: returns a list of the ip-ranges strings in the current IPBlock object
func (b *IPBlock) ToIPRangesList() []string {
```

adisos (changes requested):
a few comments

pkg/common/CanonicalIntervalSet.go
ShiriMoran: done, even though I think the code looked better before :-)

pkg/vpcmodel/externalNetwork.go
```diff
@@ -153,3 +156,22 @@ func isEntirePublicInternetRange(nodes []Node) (bool, error) {
...
+		return ""
+	}
+	// 2. Union all IPBlocks in a single one; its intervals will be the cidr blocks or ranges that should be printed, after all possible merges
+	unionBlock := &common.IPBlock{}
+	for _, ipBlock := range ipbList {
+		unionBlock = unionBlock.Union(ipBlock)
+	}
+	// 3. print a list s.t. each element contains either a single cidr or an ip range
+	return strings.Join(unionBlock.ListToPrint(), commaSepartor)
+}
```
adisos, line 177: I think this function should be moved to the file `grouping.go` (where `groupedExternalNodes` is defined).

adisos (dismissed):

Issue #131, opened 2023-08-08
grouping follow-ups


ShiriMoran:
- [ ] extend the support of ip-block grouping to `all_subnets` analysis type as well  (see issue #124)

- [ ] support grouping of ip-blocks for json format : note that the non-grouped version has more information about the connections than stateful/stateless)
- [x] grouped output issue: external nodes are printed without "Public Internet" prefix

ShiriMoran:
third bullet solved by https://github.com/np-guard/vpc-network-config-analyzer/issues/131

ShiriMoran:
first item moved to #124 which was already solved
2nd item moved to #171 
closing this one

PR #135, opened 2023-08-10
131 print public internet

ShiriMoran:


ShiriMoran:
solves 3rd bullet of https://github.com/np-guard/vpc-network-config-analyzer/issues/131

pkg/vpcmodel/grouping.go
```diff
@@ -84,7 +84,7 @@ func (g *groupedExternalNodes) Name() string {
 	if err != nil {
 		return ""
 	}
-	return toPrint
+	return "Public Internet " + toPrint
```
adisos, line 87: use a const var for this

adisos (changes requested):

Issue #124, opened 2023-08-03
grouping support for all analysis types 


adisos:
support grouping with `all_subnets`:
1. Group external addresses
2. Group subnets

not relevant to `single_subnet`

ShiriMoran:
1 above solved by #136

ShiriMoran:
2 above solved by https://github.com/np-guard/vpc-network-config-analyzer/pull/142

PR #136, opened 2023-08-14
124 all subnets grouping

ShiriMoran:


ShiriMoran:
solves #124

pkg/ibmvpc/examples/acl_testing5_oldsubnetsBased_withPGW.txt
```diff
@@ -1,11 +1,13 @@
...
 sub1-2-ky => sub1-1-ky : protocol: TCP
 sub1-2-ky => sub1-3-ky : protocol: TCP
 sub1-3-ky => sub1-1-ky : protocol: TCP
 sub1-3-ky => sub1-2-ky : protocol: TCP
-sub2-1-ky => Public Internet [8.8.8.8/32] : protocol: UDP dst-ports: 53
+sub2-1-ky => Public Internet 8.8.8.8/32 : protocol: UDP dst-ports: 53
 sub2-1-ky => sub2-2-ky : All Connections
 sub2-2-ky => sub2-1-ky : All Connections
+
+connections are stateful unless marked with *
```
adisos, line 13: we do not yet have any implemented computation of whether subnets connectivity is stateful.
We should remove that comment , and have another issue to handle stateful computation for this as well,

ShiriMoran: Opened https://github.com/np-guard/vpc-network-config-analyzer/issues/138 (next in my todos) as well as ugly WA code that will be removed once the issue is solved

zivnevo (approved):
LGTM

We have to consider if `EndpointElem` is really required in the code, given that we already have `VPCResourceIntf`. This may allow us to reduce code duplication, as `Node` also embeds `VPCResourceIntf`. This can wait for now, but maybe open an issue for that...

ShiriMoran:
> LGTM
> 
> We have to consider if `EndpointElem` is really required in the code, given that we already have `VPCResourceIntf`. This may allow us to reduce code duplication, as `Node` also embeds `VPCResourceIntf`. This can wait for now, but maybe open an issue for that...

https://github.com/np-guard/vpc-network-config-analyzer/issues/141 
Btw - my (not yet ready) next PR further uses endpoint element

Issue #127, opened 2023-08-06
More unit test for grouping


ShiriMoran:
End-to-end examples:

1. Examples in which the grouping differs due to https://github.com/np-guard/vpc-network-config-analyzer/pull/125 in the content and number of lines, and not only in having the additional '*'; an example for external addresses and one for internal ones.  Unit test?
1. Grouped external addresses results in an interval of few `cidrs` that is presented as `ip range` in output.

PR #142, opened 2023-08-17
124 all subnets grouping subnets

ShiriMoran:


ShiriMoran:
Tests for all-subnets + grouping should be added, https://github.com/np-guard/vpc-network-config-analyzer/issues/127

zivnevo (changes requested):
Some minimal tests are missing here.
Otherwise, looks good.

ShiriMoran:
> Some minimal tests are missing here. Otherwise, looks good.

I've added a very basic test of sg_testing1_new with  for all_subnets with grouping 
I'll add more tests soon https://github.com/np-guard/vpc-network-config-analyzer/issues/127

zivnevo (approved):

PR #145, opened 2023-08-23
preparation for Grouping in drawio

haim-kermany:
made some preparation for grouping in drawio:
1. created an icon for network interface, in the same foramat of all other icons.
2. that made the template code much simpler.
3. added tooltip to icons
4. made small layout changes ( NIs with FIP can not share matix cell)


zivnevo (approved):
LGTM

Issue #141, opened 2023-08-17
Is endpointElement really required


ShiriMoran:
LGTM

We have to consider if `EndpointElem` is really required in the code, given that we already have `VPCResourceIntf`. This may allow us to reduce code duplication, as `Node` also embeds `VPCResourceIntf`. This can wait for now, but maybe open an issue for that...

_Originally posted by @zivnevo in https://github.com/np-guard/vpc-network-config-analyzer/pull/136#pullrequestreview-1580452366_

adisos:
PR #222 has replaced  `EndpointElem` with `VPCResourceIntf` on connectivity structures before grouping.
The  `EndpointElem` type is currently used to represent grouped elements.

ShiriMoran:
The delta between `EndpointElem`  and `VPCResourceIntf ` is that the latter interface implements only `Name()` while the latter also implements `UID()`, `Name()`, `ZoneName()` and `Kind()`; the former functionality is non-relevant for grouped elements but required for VPC resources. Thus, the answer seems "yes",  `EndpointElem` is required

PR #146, opened 2023-08-23
144 grouping drawio

ShiriMoran:


ShiriMoran:
solves #146

pkg/vpcmodel/grouping.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 const commaSepartor = ","
+const implies = " => "
```
zivnevo, line 11: "implies" might be confusing here. I would use "connectsTo" or simply "arrow"

pkg/vpcmodel/abstractVPC.go
```diff
@@ -8,6 +8,8 @@ import (
 type VPCResourceIntf interface {
 	UID() string
 	Name() string
+	Names() (string, []string)
+	Type() string
```
zivnevo, line 12: It's unclear for the reader what's the difference between `Name()` and `Names()` and why we need both. Please add a comment

zivnevo, line 12: Where does someone get the list of available type strings to compare against? Seems like they are scattered in several files.

zivnevo (changes requested):

pkg/vpcmodel/grouping.go
ShiriMoran: done

pkg/vpcmodel/abstractVPC.go
ShiriMoran: done

ShiriMoran: Added a comment as to why this is needed. Not sure I understand the 2nd sentence

pkg/vpcmodel/abstractVPC.go
```diff
@@ -8,6 +8,8 @@ import (
 type VPCResourceIntf interface {
 	UID() string
 	Name() string
+	Names() (string, []string)
```
haim-kermany, line 12: having both Name() and Names() is strange. maybe change it to DisplayNames()

pkg/vpcmodel/grouping.go
```diff
@@ -85,18 +90,44 @@ func (g *groupedEndpointsElems) Name() string {
...
 }
 
+func (g *groupedEndpointsElems) Names() (string, []string) {
+	names := func(ep EndpointElem) string {
+		myName, _ := ep.Names()
+		return myName
+	}
+	namesToPrint := endpointElemToPrint(*g, names)
+	return strings.Join(namesToPrint, commaSepartor), namesToPrint // todo Haim: let me know if you want something shorter at the first index
+}
```
haim-kermany, line 101: yes, it should be 20 chars max

pkg/vpcmodel/grouping.go
```diff
@@ -288,6 +319,20 @@ func (g *GroupConnLines) String() string {
...
+			linesStr[i] = line.Src.Name() + implies + line.Dst.Name() // do not print connection string
+		} else {
+			linesStr[i] = line.String()
+		}
+	}
+	sort.Strings(linesStr)
+	return strings.Join(linesStr, "\n")
+}
+
 // StringTmpWA ToDo: tmp WA until https://github.com/np-guard/vpc-network-config-analyzer/issues/138.
```
haim-kermany, line 337: not shore what StringForDrawio() is for

pkg/vpcmodel/abstractVPC.go
zivnevo: Is `Type()` meant just to be printed as-is to the drawio file, or is it going to be used programmatically to check the type of a give VPC resource? If the latter is true, how do I know what is the correct string to check against? Do I have to take a look at the code and copy the same type string to my code?

haim-kermany (changes requested):

pkg/vpcmodel/abstractVPC.go
ShiriMoran: I guess the latter is true, but this is a question to @haim-kermany: if the current interface which outputs string (I did it quick and dirty, true) is not good enough, please suggest the required interface. Do we need to introduce enums?

haim-kermany: Type() is needed to distinguish between kind of grouping (i.e. endpoint).
each grouping is implemented differently in the drawio.
I do not know yet what kind of  grouping will we have.
 but the current implementation of returning strings is problematic, since some of the stings are defined in the ibmvpc pkg
.

pkg/vpcmodel/abstractVPC.go
ShiriMoran: np

pkg/vpcmodel/grouping.go
ShiriMoran: Cool, but what should the short name be then? I've put a placeholder

pkg/vpcmodel/grouping.go
ShiriMoran: `StringForDrawio()` is supposed to be the `nice connection name, without "protocol:"` you asked for, but now I see it does something else; will fix.

pkg/vpcmodel/abstractVPC.go
ShiriMoran: I believe the next comment by Haim and my fix also resolves this one

ShiriMoran: @haim-kermany your next comment is about this. I fixed by using the existing `Kind()` as you ask for - however, these uses strings defined in `ibmvpc` which you can not import.  
Perhaps move these constants to some shared library, also the new constants defined in `grouping.go`?  @zivnevo

pkg/ibmvpc/vpc.go
```diff
@@ -61,11 +62,17 @@ func (ni *NetworkInterface) VsiName() string {
...
 func (ni *NetworkInterface) Name() string {
 	return getNodeName(ni.vsi, ni.address)
 }
+func (ni *NetworkInterface) Names() (string, []string) {
+	return ni.Name(), []string{ni.Name()}
+}
+func (ni *NetworkInterface) Type() string {
+	return networkInterface
+}
 func (ni *NetworkInterface) Details() []string {
```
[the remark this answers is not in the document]
ShiriMoran: See Ziv's comment above and discussion there

pkg/vpcmodel/abstractVPC.go
zivnevo: @haim-kermany , where exactly do you have to use these Kinds? Currently they are only used for preparing the drawio  TreeNode model in `pkg\vpcmodel\drawioOutput.go` which is part of `vpcmodel`. The new Kinds in `grouping.go` are also in `vpcmodel` so I see no circular dependencies. Would you like the new grouped types to implement the `VPCResourceIntf` interface, and especially `DetailsMap()` where you usually take the Kind from?

pkg/vpcmodel/grouping.go
ShiriMoran: Looking at the code I think more details are needed @haim-kermany.
1. Printing the "connection":  is done by the following function, and is the same with and without `grouping`. Perhaps what is required is a dedicated printing function for `drawio` which I can use also for `grouping`; how do you handle the same issue for the non grouping case?
`func (conn *ConnectionSet) String() string {`

2. What exactly do you mean by `nice connection name, without "protocol:`?
 This is where "protocol:" comes from. Note that there are other stuff there other than "protocol"
 Please define accurately what is it you want to print in connections. 
```
 func getDimensionStr(dimValue *CanonicalIntervalSet, dim Dimension) string {
	domainValues := getDimensionDomain(dim)
	if dimValue.Equal(*domainValues) {
		// avoid adding dimension str on full dimension values
		return ""
	}
	switch dim {
	case protocol:
		pList := []string{}
		for p := minProtocol; p <= maxProtocol; p++ {
			if dimValue.Contains(p) {
				pList = append(pList, getProtocolStr(p))
			}
		}
		return "protocol: " + strings.Join(pList, ",")
	case srcPort:
		return "src-ports: " + dimValue.String()
	case dstPort:
		return "dst-ports: " + dimValue.String()
	case icmpType:
		return "icmp-type: " + dimValue.String()
	case icmpCode:
		return "icmp-code: " + dimValue.String()
	}
	return ""
}
```

pkg/vpcmodel/abstractVPC.go
haim-kermany: I do not like DetailsMap(). it is not clear what are the keys there. I used because I do not want anything else.
Idealy,  if VPCResourceIntf could give me a better interface than DetailsMap(), it would be better.
In grouping, I have another problem. i need to know the type of EndpointElem( which is not only VPCResourceIntf )

zivnevo: @haim-kermany , maybe open a new issue about how you suggest the `abstractVPC -> drawio` interface to look like. It's not to late to refactor. Maybe we can do it together with issue #141 .
For this work, I suggest to still use the `DetailsMap()`, and maybe add new keys to it to reflect the type of `EndpointElem`.

haim-kermany: I just relized we have muke kind of grouping that I thought.
I will be much wizer after I know what kind of grouping do we have, and if/how we will present them
will all should discuss with Adi and keren

Issue #98, opened 2023-07-06
improve grouping with self-loops


adisos:
```suggestion
vsi3a-ky[10.240.30.5],vsi3b-ky[10.240.30.6],vsi3c-ky[10.240.30.4] => vsi3a-ky[10.240.30.5],vsi3b-ky[10.240.30.6],vsi3c-ky[10.240.30.4] : All Connections
vsi3a-ky[10.240.30.5],vsi3b-ky[10.240.30.6],vsi3c-ky[10.240.30.4] => vsi1-ky[10.240.10.4] : All Connections
```
This requires adding a self loop to each VSI, run the grouping algorithm, then remove left-over self loops.
Can be handled in a future PR.

_Originally posted by @zivnevo in https://github.com/np-guard/vpc-network-config-analyzer/pull/92#discussion_r1253989708_

ShiriMoran:
> ```
> This requires adding a self loop to each VSI, run the grouping algorithm, then remove left-over self loops.
> Can be handled in a future PR.
> 
> _Originally posted by @zivnevo in https://github.com/np-guard/vpc-network-config-analyzer/pull/92#discussion_r1253989708_
>             
> ```

@zivnevo adding a self loop before running the algorithm seems wrong.
Consider subnets `a,b,c,d` with the following connections:
```
a => b
a => c
d => c
d => b
```
With the current algorithm, after grouping src we'll get
```
a => b,c
d => b,c
```
and after grouping dst on the result we'll get
`a,d => b,c`

but if we add self loops before running the alg after grouping src we'll get:
```
a => a,b,c
d => b,c,d
```
and grouping dst will group nothing.

I'm working on a different solution.

ShiriMoran:
Actually I'm afraid what is really required is to check all the possible combinations of with and without self loops, which is not reasonable complexity rise. I have another solution in mind, if you wish we can discuss or I can put it here before I write the code.

ShiriMoran:
Proposed algorithm:
* execute the current grouping algorithm for grouping, assume w.l.o.g we grouped destinations 
* Merge lines whose distance is zero until reaching a fix point.

The distance between a line 1 with src `a_1` and dst `b_2`  and line 2 with src `a_2` and distance `b_2` is defined as follows:

ShiriMoran:
solved by #152

PR #152, opened 2023-09-05
98 grouping with self loops

ShiriMoran:
solves #98


adisos (commented):
Added a few initial comments

pkg/vpcmodel/grouping.go
```diff
@@ -204,6 +205,20 @@ func (g *GroupConnLines) groupExternalAddressesForSubnets() {
 	g.GroupedLines = res
 }
 
+// aux func, returns true iff the EndpointElem is Node if grouping vsis or NodeSet if grouping subnets
+func isInetrnalOfRequiredType(ep EndpointElem, groupVsi bool) bool {
```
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/groupingSelfLoop.go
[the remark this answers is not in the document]
ShiriMoran: Done
I have spell checker in `goland`.

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,330 @@
+package vpcmodel
+
+import (
+	"strings"
+)
+
+// extends grouping by considering self loops as don't care https://github.com/np-guard/vpc-network-config-analyzer/issues/98
+// e.g. a => b,c   b => a, c   and   c => a,b   is actually a clique a,b,c => a,b,c
+// a => b,c, d => b,c can be presented in one line as  a, d => b,c
```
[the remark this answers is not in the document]
ShiriMoran: typo. corrected.

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,330 @@
...
+
+// After the basic grouping, which is of worst time complexity O(n^2), we optimize grouping treating self loops as don't care.
+// Intuitively, we check if two GroupedConnLine can be merged treating self loops as don't care
+// 1. groupsToBeMerged find couples of GroupedConnLine that should be merged using the alg below
+// mergeSelfLoops merges the groupsToBeMerged:
+// 2. It creates sets of GroupedConnLine s.t. each set should be merged
```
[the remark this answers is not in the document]
ShiriMoran: I find the term equivalence groups confusing here. Wouldn't it be better to say:
"note that the "should be merged" is an equivalence relation" ?

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,330 @@
...
+				toMergeCouples = append(toMergeCouples, toMerge)
+			}
+		}
+	}
+	return toMergeCouples
+}
+
+// a group is candidate to be merged only if it has only internal nodes
+func relevantKeysToCompare(groupingSrcOrDst map[string][]*GroupedConnLine) (sortedKeys []string) {
+	sortedKeys = make([]string, 0, len(groupingSrcOrDst))
```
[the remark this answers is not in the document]
ShiriMoran: I already did yesterday and apparently forget to push

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,330 @@
...
+		// 	going over all couples of items: merging them if they differ only in self loop element
+		// mergeCandidates of a singleton 'key' are all lines in which the group contains 'key'
+		//  if key is not a singleton then mergeCandidates will be empty
+		mergeCandidates, ok := keyToMergeCandidates[key]
+		if !ok {
+			continue
+		}
+		for candidate := range mergeCandidates {
+			candidateLines := groupingSrcOrDst[candidate]
+			// delta between outerKeyEndPointElements to innerKeyEndPointElements must be 0
```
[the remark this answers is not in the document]
ShiriMoran: sure. This is old comment that somehow I missed

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,330 @@
...
+			mergeGroups := deltaBetweenGroupedConnLines(srcGrouping, keyLines, candidateLines, setsToGroup[key], setsToGroup[candidate])
+			// delta between the keyLines is 0 - merge keyLines
+			if mergeGroups {
+				var toMerge = [2]string{key, candidate}
+				toMergeCouples = append(toMergeCouples, toMerge)
```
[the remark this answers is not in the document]
ShiriMoran: sure

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,330 @@
...
+//  1. They are of the same connection
+//  2. If vsis, of the same subnet
+//  3. The src/dst is a singelton contained in the dst/src
+//     in one path on groupingSrcOrDst we prepare a map between each key to the keys that are candidate to be merged with it.
+//     Before the grouping there are at most O(n^20) lines of src -> dst
+//     The last condition implies that each original src -> dst (where src and dst are a single endpoint) can induce a single
+//     candidate (at most), and each singelton key have at most n candidates. Hence there are at most
+//     O(n^3) merge candidate, which implies O(n^3) time complexity of groupsToBeMerged
+func (g *GroupConnLines) mergeCandidates(groupingSrcOrDst map[string][]*GroupedConnLine, srcGrouping bool,
+	keyToGroupedSets map[string]map[string]struct{}, sortedKeys []string) map[string]map[string]struct{} {
```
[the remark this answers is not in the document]
ShiriMoran: sure

pkg/vpcmodel/grouping_test.go
```diff
@@ -140,3 +140,152 @@ func TestGroupingPhase2(t *testing.T) {
...
+	res1.AllowedConnsCombined.updateAllowedConnsMap(res.Nodes[0], res.Nodes[2], common.NewConnectionSet(true))
+	res1.AllowedConnsCombined.updateAllowedConnsMap(res.Nodes[1], res.Nodes[0], common.NewConnectionSet(true))
+	res1.AllowedConnsCombined.updateAllowedConnsMap(res.Nodes[1], res.Nodes[2], common.NewConnectionSet(true))
+	res1.AllowedConnsCombined.updateAllowedConnsMap(res.Nodes[2], res.Nodes[1], common.NewConnectionSet(true))
+	res1.AllowedConnsCombined.updateAllowedConnsMap(res.Nodes[2], res.Nodes[0], common.NewConnectionSet(true))
+
+	return res, res1
+}
+
+func TestGroupingPhase3(t *testing.T) {
```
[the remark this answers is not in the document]
ShiriMoran: renamed to reflect the tests semantic

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,335 @@
...
+	keyToMergeCandidates := make(map[string]map[string]struct{})
+	// 2. in each bucket finds for each key the candidates to be merged, in two stages
+	for _, keysInBucket := range bucketToKeys {
+		singletonsInBucket := make(map[string]string)
+		//    2.1 for a group g_1 s.t. the non-grouped src/dst is a singleton,
+		//        all groups in which the grouped dst/src contains the singleton
+		//        2.1.1 finds for each bucket all singletons
+		for key := range keysInBucket {
+			lines := groupingSrcOrDst[key]
+			elemsInkey := elemInKeys(!srcGrouping, *lines[0])
```
adisos, line 117: rename `elemsInkey` to `elemsInKey`

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,335 @@
...
+// a couple of []*GroupedConnLine is candidate to be merged only if:
+//  1. They are of the same connection
+//  2. If vsis, of the same subnet
+//  3. The src/dst is a singleton contained in the dst/src
+//     in one path on groupingSrcOrDst we prepare a map between each key to the keys that are candidate to be merged with it.
+//     Before the grouping there are at most O(n^20) lines of src -> dst
+//     The last condition implies that each original src -> dst (where src and dst are a single endpoint) can induce a single
+//     candidate (at most), and each singleton key have at most n candidates. Hence, there are at most
+//     O(n^3) merge candidate, which implies O(n^3) time complexity of groupsToBeMerged
+func (g *GroupConnLines) mergeCandidates(groupingSrcOrDst map[string][]*GroupedConnLine, srcGrouping bool,
```
adisos, line 93: function name is confusing, this function computes candidates to be merged later, and does not really merge.

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,335 @@
...
+			newGroupedConnLine = append(newGroupedConnLine, &GroupedConnLine{epInLineValue, &epsInNewKey, conn})
+		} else {
+			newGroupedConnLine = append(newGroupedConnLine, &GroupedConnLine{&epsInNewKey, epInLineValue, conn})
+		}
+	}
+	srcsOrDstsInNewKeySlice := make([]string, 0)
+	for item := range namesInNewKey {
+		srcsOrDstsInNewKeySlice = append(srcsOrDstsInNewKeySlice, item)
+	}
+	newKey = strings.Join(srcsOrDstsInNewKeySlice, commaSeparator) + conn
```
adisos, line 333: The format of the key here is a bit different from the original format (see line 233 in `grouping.go`). 
Consider adding a function to handle the key's format uniquely in all places.

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,335 @@
...
+//
+// The distance between two GroupedConnLine is defined as following:
+// Let l_1 be a line with source s_1 and dest d_1 and let l_2 be a line with source s_2 and dest d_2.
+// l_1 / l_2 is the subnets in d_1 that are not in d_2 minus the single subnet in s_1 if |s_1| = 1
+//
+// The distance between lines l_1 and l_2 is l_1 / l_2 union l_2 / l_2
+//
+// claim: if the distance between line l_1 and l_2 is empty and
+//		 the distance between lines l_2 and l_3 is zero
+//		 then so is the distance between l_1 and l_3
```
adisos, line 31: be consistent with distance term `zero` or `empty`

pkg/vpcmodel/groupingSelfLoop.go
ShiriMoran: > rename `elemsInkey` to `elemsInKey`

These are identical. Perhaps you meant something else?

pkg/vpcmodel/groupingSelfLoop.go
ShiriMoran: renamed to `findMergeCandidates`

pkg/vpcmodel/groupingSelfLoop.go
adisos: not identical, note the capital `k`

pkg/vpcmodel/groupingSelfLoop.go
ShiriMoran: Seems line 233 in `grouping.go` has shifted. To be on the safe side, please refer to the relevant code (not line)

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,335 @@
...
+//			note that the "should be merged" is an equivalence relation
+// 3. It merges them
+
+// alg: GroupedConnLine to be merged:
+// GroupedConnLine whose distance is an empty set should be merged;
+// the claim below guarantees that the result is coherent
+//
+// The distance between two GroupedConnLine is defined as following:
+// Let l_1 be a line with source s_1 and dest d_1 and let l_2 be a line with source s_2 and dest d_2.
+// l_1 / l_2 is the subnets in d_1 that are not in d_2 minus the single subnet in s_1 if |s_1| = 1
```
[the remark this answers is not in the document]
ShiriMoran: right. fixed.

pkg/vpcmodel/groupingSelfLoop.go
ShiriMoran: Walla

pkg/vpcmodel/groupingSelfLoop.go
adisos: the line I referred to: `key := dstOrSrc.Name() + ";" + line.Conn`

pkg/vpcmodel/groupingSelfLoop.go
ShiriMoran: right. fixed

adisos (commented):
General comment: please add documentation for individual functions about their input args and return values, and what they represent, especially for non-trivial functions (such as `deltaBetweenGroupedConnLines` , `setMinusSet`... ).

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,333 @@
...
+	return strings.Split(srcOrDst.Name(), commaSeparator)
+}
+
+func setMinusSet(srcGrouping bool, groupedConnLine GroupedConnLine, set1, set2 map[string]struct{}) map[string]struct{} {
+	minusResult := make(map[string]struct{})
+	for k := range set1 {
+		if _, ok := set2[k]; !ok {
+			minusResult[k] = struct{}{}
+		}
+	}
```
adisos, line 218: A simpler type for representing a set can be `map[string]bool` instead of `map[string]struct{}` (when a key is added it is mapped to `true`) so that to check if `k` is in the set you simply use expression `if set[k] { ...} `  instead of 
```
if _, ok := set[k] ; ok{

}
```
consider to simplify in relevant places in the code

ShiriMoran: I am familiar with this representation of sets; is it indeed simpler? because `Bool` can be `false` or `true`; as such, checking for existence in the map is not enough and the content of the map also needs to be checked. In `map[string]struct{}` it is clear that the value of the map is meaningless.

adisos: Any value in the map is always mapped to `True`, so that the expression `set[k]` is True iff the key `k` is in the map

ShiriMoran: I disagree here. I think the `Bool` in this context is confusing in the sense that one may assume that it can also be mapped to false, while `struct{}` clearly implies that the value is meaningless.

adisos: ok, let's keep as is now

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,333 @@
...
+		//  is there a different line s.t. the keyLines were not merged only due to self loops?
+		// 	going over all couples of items: merging them if they differ only in self loop element
+		// findMergeCandidates of a singleton 'key' are all lines in which the group contains 'key'
+		//  if key is not a singleton then findMergeCandidates will be empty
+		mergeCandidates, ok := keyToMergeCandidates[key]
+		if !ok {
+			continue
+		}
+		for candidate := range mergeCandidates {
+			candidateLines := groupingSrcOrDst[candidate]
```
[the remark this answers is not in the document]
ShiriMoran: Now this will be the case only if both `x` and `y` are singeltons. Not sure its worth the trouble in this case. In fact, even complexity wise its not worth while. Its no longer a matrix.

pkg/vpcmodel/groupingSelfLoop.go
```diff
@@ -0,0 +1,335 @@
...
+//     O(n^3) merge candidate, which implies O(n^3) time complexity of groupsToBeMerged
+func (g *GroupConnLines) mergeCandidates(groupingSrcOrDst map[string][]*GroupedConnLine, srcGrouping bool,
+	keyToGroupedSets map[string]map[string]struct{}, relevantKeys []string) map[string]map[string]struct{} {
+	// 1. Create buckets for each connection + vsi's subnet if vsi; merge candidates are within each bucket
+	bucketToKeys := make(map[string]map[string]struct{})
+	for _, key := range relevantKeys {
+		lines := groupingSrcOrDst[key]
+		bucket := lines[0].Conn
+		bucket = g.addVsiSubnetToBucket(lines[0].Src, bucket)
+		bucket = g.addVsiSubnetToBucket(lines[0].Dst, bucket)
```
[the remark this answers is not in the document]
ShiriMoran: agreed. Done

pkg/vpcmodel/groupingSelfLoop.go
ShiriMoran: I'm not sure I am following. Perhaps we should take it offline

ShiriMoran: Groups are merged only after it was already verified they have the same connection. Hence it makes no sense to add the connection to the key. This is different than the lines you quota from `grouping.go`

adisos: I meant that you don't add `;` between the different parts in the key. In you version it is `strings.Join(srcsOrDstsInNewKeySlice, commaSeparator) + conn` , and in the line I was referring to, the `;` is added before the `conn` part.
I think the format should be the same.

ShiriMoran: Ahh, now I understand.

adisos (approved):

Issue #138, opened 2023-08-15
Add statefull analysis to "all subnets" mode


ShiriMoran:

ShiriMoran:
Currently the mode lacks this check

ShiriMoran:
closed by #154

Issue #56, opened 2023-05-11
differentiate between stateful/stateless connections in the analysis


adisos:
**SG**: Rules are stateful. To have stateful rules means when you create a rule that allows traffic one way, the rule automatically permits reverse traffic. For example, if you create a rule to allow inbound TCP traffic on port 80, the rule also allows replying outbound TCP traffic on port 80 back to the originating host, without the need for another rule.

**NACL**: An Access Control List (ACL) can manage (that is, it can allow or deny) inbound and outbound traffic for a subnet. An ACL is stateless, which means that inbound and outbound rules must be specified separately and explicitly. Each ACL consists of rules, based on a source IP, source port, destination IP, destination port, and protocol.

In the connectivity output, a connection `A -> B` is _stateful (allows bidrectional flow)_ if both SG and NACL (of A and B) allow connection (ingress and egress) from A to B , AND if NACL (of A and B) allow connection (ingress and egress) from B to A .

PR #154, opened 2023-09-11
138 statefull all subnets

ShiriMoran:
solves #138


pkg/vpcmodel/subnetsConnectivity.go
```diff
@@ -252,14 +255,45 @@ func (v *VPCsubnetConnectivity) computeAllowedConnsCombined() error {
...
+			default:
+			}
+			conns.IsStateful = common.StatefulFalse
+			if otherDirectionConn == nil {
+				continue
+			}
+			stateful, err := conns.ContainedIn(otherDirectionConn)
+			if err != nil {
+				return err
+			}
```
adisos, line 284: why checking for a containment in specific direction? if the intersection between `conns` and `otherDirectionConn` is not empty, isn't it sufficient to be stateful?
Take a look at `nodesConnectivity.go` lines 237-247. Should also use `SwitchSrcDstPorts()` before checking the intersection here as well (as explained in https://github.com/np-guard/vpc-network-config-analyzer/issues/56#issuecomment-1586154619 ).

pkg/vpcmodel/subnetsConnectivity.go
```diff
@@ -252,14 +255,45 @@ func (v *VPCsubnetConnectivity) computeAllowedConnsCombined() error {
...
+		for dst, conns := range endpointConns {
+			if conns.IsEmpty() {
+				continue
+			}
+			dstObj := v.cloudConfig.NameToResource[dst.Name()]
+			var otherDirectionConn *common.ConnectionSet = nil
+			switch dstObj.(type) {
+			case NodeSet:
+				otherDirectionConn = v.AllowedConnsCombined[dst][src]
+			case *ExternalNetwork:
```
adisos, line 272: why do we expect `dstObj`  as returned from `NameToResource` to be `ExternalNetwork` ?

pkg/vpcmodel/subnetsConnectivity.go
```diff
@@ -252,14 +255,45 @@ func (v *VPCsubnetConnectivity) computeAllowedConnsCombined() error {
...
+}
 
+func (v *VPCsubnetConnectivity) computeStatefulConnections() error {
+	for src, endpointConns := range v.AllowedConnsCombined {
+		for dst, conns := range endpointConns {
+			if conns.IsEmpty() {
+				continue
+			}
+			dstObj := v.cloudConfig.NameToResource[dst.Name()]
+			var otherDirectionConn *common.ConnectionSet = nil
```
adisos, line 268: initialized as `nil` by default in `var` declaration?

pkg/vpcmodel/subnetsConnectivity.go
```diff
@@ -252,14 +255,45 @@ func (v *VPCsubnetConnectivity) computeAllowedConnsCombined() error {
...
+			switch dstObj.(type) {
+			case NodeSet:
+				otherDirectionConn = v.AllowedConnsCombined[dst][src]
+			case *ExternalNetwork:
+				// subnet to external node is stateful if the subnet's nacl allows egress from that node
+				otherDirectionConn = v.AllowedConns[src].EgressAllowedConns[dst]
```
adisos, line 274: I do not understand this line: from the lines above, we know that there is allowed conn `src => dst : conns`, and in this case of the switch, `dst` is external network. 
So for it to be considered `stateful`, the subnet `src` should be able to receive responses from this external network, so there should be an allowed connection  from `dst` to `src`. why checking `EgressAllowedConns` to `dst`? 
Why not checking there is an allowed combined connection from `dst` to `src`?

pkg/vpcmodel/subnetsConnectivity.go
ShiriMoran: There are two comments here: one about using `SwitchSrcDstPorts()` - sure, I'll take a look and use.
Not sure I understand the other comment. Does it refer to the following?
```
case *ExternalNetwork:
				// subnet to external node is stateful if the subnet's nacl allows egress from that node
				otherDirectionConn = v.AllowedConns[src].EgressAllowedConns[dst]
```

pkg/vpcmodel/subnetsConnectivity.go
ShiriMoran: Because there is no actual connection in this direction, only for stateful. Hence this direction will not be part of the combined connections.

pkg/vpcmodel/subnetsConnectivity.go
ShiriMoran: Why not? Seems we are not on the same page -  perhaps we should take it offline

pkg/vpcmodel/subnetsConnectivity.go
ShiriMoran: right. fixed

pkg/vpcmodel/subnetsConnectivity.go
ShiriMoran: I ended not adding `SwitchSrcDstPorts()` - since the ingress external nodes is missing here I think it is not really helpful here.

pkg/vpcmodel/subnetsConnectivity.go
ShiriMoran: Fixed to Ingress, of course

pkg/vpcmodel/subnetsConnectivity.go
adisos: why is `SwitchSrcDstPorts()` related to `ingress external nodes` ?
The connection set from `otherDirection` should have the ports flipped before intersection with `conns`.

ShiriMoran:
@adisos  SwitchSrcDstPorts() is not defined for ICMP  protocol, ICMP protocol (which are not TCP or UDP)  connections are identified as not statefull - in `acl_testing5subnetsBased_withPGW`
Why isn't  `SwitchSrcDstPorts()` defined for ICMP protocol ?
The connection seems stateful by nature https://www.fortinet.com/resources/cyberglossary/internet-control-message-protocol-icmp

adisos:
> @adisos SwitchSrcDstPorts() is not defined for ICMP protocol, ICMP protocol (which are not TCP or UDP) connections are identified as not statefull - in `acl_testing5subnetsBased_withPGW` Why isn't `SwitchSrcDstPorts()` defined for ICMP protocol ? The connection seems stateful by nature https://www.fortinet.com/resources/cyberglossary/internet-control-message-protocol-icmp . Is there an open issue in this regard? anyways, this potential issue seems non-related to the current one

opened issue #170

adisos (approved):

PR #176, opened 2023-09-27
Support grouping in drawio for external nodes

haim-kermany:


pkg/vpcmodel/grouping.go
```diff
@@ -75,6 +75,7 @@ type GroupConnLines struct {
 // EndpointElem can be Node(networkInterface) / groupedExternalNodes / groupedNetworkInterfaces
```
adisos, line 75:
```suggestion
// EndpointElem can be Node(networkInterface) / groupedExternalNodes / groupedEndpointsElems / NodeSet(subnet)
```

pkg/vpcmodel/drawioGenerator.go
```diff
@@ -54,3 +55,29 @@ func (gen *DrawioGenerator) TreeNode(res DrawioResourceIntf) drawio.TreeNodeInte
...
+	if len(*g) == 1 {
+		return gen.TreeNode((*g)[0])
+	}
+	tooltip := []string{}
+	for _, n := range *g {
+		tooltip = append(tooltip, n.(*ExternalNetwork).Cidr())
+	}
+	name := "multi cidr"
+	if all, _ := isEntirePublicInternetRange(*g); all {
+		name = "all ranges"
```
adisos, line 74: consider renaming to "Public Internet"

pkg/vpcmodel/drawioGenerator.go
```diff
@@ -54,3 +55,29 @@ func (gen *DrawioGenerator) TreeNode(res DrawioResourceIntf) drawio.TreeNodeInte
...
+
+func (g *groupedExternalNodes) GenerateDrawioTreeNode(gen *DrawioGenerator) drawio.TreeNodeInterface {
+	if len(*g) == 1 {
+		return gen.TreeNode((*g)[0])
+	}
+	tooltip := []string{}
+	for _, n := range *g {
+		tooltip = append(tooltip, n.(*ExternalNetwork).Cidr())
+	}
+	name := "multi cidr"
```
adisos, line 72: Consider renaming to "various IP ranges" / "CIDR list" /... ?

pkg/ibmvpc/ibmDrawioGenerator.go
```diff
@@ -5,6 +5,17 @@ import (
...
+func (v *VPC) IsExternal() bool                  { return false }
+func (z *Zone) IsExternal() bool                 { return false }
+func (s *Subnet) IsExternal() bool               { return false }
+func (sgl *SecurityGroupLayer) IsExternal() bool { return false }
+func (nl *NaclLayer) IsExternal() bool           { return false }
+func (n *IKSNode) IsExternal() bool              { return false }
+func (ni *NetworkInterface) IsExternal() bool    { return false }
+func (v *Vsi) IsExternal() bool                  { return false }
+func (pgw *PublicGateway) IsExternal() bool      { return false }
+func (fip *FloatingIP) IsExternal() bool         { return false }
```
adisos, line 17: An alternative is to add the method `IsExternal() bool` to the `VPCResource` struct in `vpcmodel/abstractVPC.go` such that it just returns `true` there , and also change the type `ExternalNetwork` in `vpcmodel/externalNetwork.go` so that it will no longer contain `VPCResource` (and add its methods to implement the `VPCResourceIntf` ).

pkg/vpcmodel/grouping.go
```diff
@@ -86,6 +87,13 @@ type GroupedConnLine struct {
 func (g *GroupedConnLine) String() string {
 	return g.Src.Name() + " => " + g.Dst.Name() + " : " + g.Conn
 }
+func (g *GroupedConnLine) ConnLabel() string {
+	// todo - this info can be found in the conn struct.
+	if strings.Contains(g.Conn, "All Connections") {
+		return ""
+	}
+	return g.Conn
```
adisos, line 95: can be added as a function exported by`common/connectionset.go` that checks if a connection string represents all connections or not. (so that won't need to duplicate the hard-coded str "All Connections").

pkg/vpcmodel/grouping.go
haim-kermany: done

pkg/vpcmodel/drawioGenerator.go
haim-kermany: done

pkg/vpcmodel/drawioGenerator.go
haim-kermany: done

pkg/ibmvpc/ibmDrawioGenerator.go
haim-kermany: done

pkg/vpcmodel/grouping.go
haim-kermany: done

pkg/common/connectionset.go
```diff
@@ -50,6 +50,13 @@ const (
 	ICMP
 )
 
+const (
+	AllConnections = "All Connections"
+	NoConnections  = "No Connections"
+)
+
+func IsAllConnections(conn string) bool { return strings.Contains(conn, AllConnections) }
```
adisos, line 58:
```suggestion
// IsAllConnections returns whether the input connection string represents all connection (true) or not (false)
func IsAllConnections(conn string) bool { return strings.Contains(conn, AllConnections) }
```

pkg/ibmvpc/ibmDrawioGenerator.go
```diff
@@ -10,6 +10,7 @@ func (v *VPC) GenerateDrawioTreeNode(gen *vpcmodel.DrawioGenerator) drawio.TreeN
 	return drawio.NewVpcTreeNode(gen.Cloud(), v.Name())
 }
 
+func (z *Zone) IsExternal() bool { return false }
```
adisos, line 13: this should be where the other methods of `Zone` are defined

pkg/vpcmodel/drawioGenerator.go
```diff
@@ -54,3 +55,30 @@ func (gen *DrawioGenerator) TreeNode(res DrawioResourceIntf) drawio.TreeNodeInte
...
+	}
+	tn := drawio.NewInternetTreeNode(gen.PublicNetwork(), name)
+	tn.SetTooltip(tooltip)
+	return tn
+}
+
+func (n *VPCResource) IsExternal() bool           { return false }
+func (exn *ExternalNetwork) IsExternal() bool     { return true }
+func (g *groupedEndpointsElems) IsExternal() bool { return false }
+func (g *groupedExternalNodes) IsExternal() bool  { return true }
```
adisos, line 84: Move these methods to the places where these types and their methods are defined

pkg/vpcmodel/externalNetwork.go
```diff
@@ -36,17 +36,19 @@ func getPublicInternetAddressList() []string {
 
 // ExternalNetwork implements Node interface
 type ExternalNetwork struct {
-	VPCResource
```
adisos, line 39: add the field `ResourceType`

pkg/vpcmodel/externalNetwork.go
```diff
@@ -36,17 +36,19 @@ func getPublicInternetAddressList() []string {
...
 	CidrStr          string
 	isPublicInternet bool
 }
 
+func (exn *ExternalNetwork) UID() string      { return "" }
+func (exn *ExternalNetwork) ZoneName() string { return "" }
+
```
adisos, line 45: should also have method `Kind()`

pkg/vpcmodel/externalNetwork.go
```diff
@@ -36,17 +36,19 @@ func getPublicInternetAddressList() []string {
...
 func (exn *ExternalNetwork) Cidr() string {
 	return exn.CidrStr
 }
 
 func (exn *ExternalNetwork) Name() string {
-	return exn.ResourceType + " [" + exn.CidrStr + "]"
+	return publicInternetNodeName + " [" + exn.CidrStr + "]"
```
adisos, line 51: use the field `ResourceType`

pkg/vpcmodel/externalNetwork.go
```diff
@@ -102,7 +104,6 @@ func newExternalNode(isPublicInternet bool, ipb *common.IPBlock) (Node, error) {
 	}
 	cidr := ipb.ToCidrList()[0]
 	return &ExternalNetwork{
-		VPCResource:      VPCResource{ResourceType: publicInternetNodeName},
```
adisos, line 105: instead of deleting, set:  `resourceType: publicInternetNodeName`

pkg/common/connectionset.go
haim-kermany: done

pkg/ibmvpc/ibmDrawioGenerator.go
haim-kermany: IsExternal() and  GenerateDrawioTreeNode() are the methods in DrawioResourceIntf.

 since we decided to separate all GenerateDrawioTreeNode() from their structs, and implement them together in one file. I thought put IsExternal() in the same file
 what do you think?

pkg/vpcmodel/drawioGenerator.go
haim-kermany: same

pkg/vpcmodel/externalNetwork.go
haim-kermany: done

pkg/vpcmodel/externalNetwork.go
haim-kermany: already implemented, line 70

pkg/vpcmodel/externalNetwork.go
haim-kermany: done

pkg/vpcmodel/externalNetwork.go
haim-kermany: done

pkg/ibmvpc/ibmDrawioGenerator.go
adisos: I think `IsExternal()` is not as related to `drawio` as the  `GenerateDrawioTreeNode` method.. Initially I thought it would better fit the `VPCResourceIntf` , but realized it is also required for grouped elements, and thus has to be in `DrawioResourceIntf`.  [though we can add this method to both `VPCResourceIntf` and `EndpointElem` interfaces instead?]

So I thought that at least this method belongs to where all the other "regular" methods are defined for each type. (it's not as unique as the special drawio functionality, and can be used for other cases).

haim-kermany: since the drawio code handle objects of DrawioResourceIntf, If I will move IsExternal() to VPCResourceIntf and EndpointElem, I wont be able to use it.
I will moving the methods to the files as you suggested, but keep it at DrawioResourceIntf.

haim-kermany: done

pkg/vpcmodel/drawioGenerator.go
haim-kermany: done

adisos (approved):

Issue #144, opened 2023-08-22
grouping api for drawio


haim-kermany:
for Endpoint Iterface please add:
1. Names() (string, []strings) - will return the the name, without "public Network", and detailed list of names
2.   an interface to get the type.  please notice that the actual type can be from the ibmvpc pkg, so I can not read the type in my code 

for GroupedConnLine:
1.  does the connection allowing all connections
2. nice connection name, without "protocol:"

haim-kermany:
one more important thing:
please do not allow two different endpoints representing the same set of nodes

ShiriMoran:
Item 2 for Endpoint interface: why not use `type switches`? this seems to be what you need, and will get you the same answer I can produce

haim-kermany:
1. grouping code is at vpcmodel package. also the drawiooutput code is at vpcmodel package. however, the types of EndpointElem are from ibmvpc package. vpcmodel can not import ibmvpc
2. yes, I need to know about connections between vsis in the subnet

ShiriMoran:
> for Endpoint Iterface please add:
> 
>     1. Names() (string, []strings) - will return the the name, without "public Network", and detailed list of names
Why do you need the additional string? This is what Name() already returns
How about: 
`Names() []strings` will return a detailed list of names,  without "public Network" (if the network interface presents a single element then the array will be of length 1)

haim-kermany:
the string should be a short representing label, to put under the icon. Name() can return a long string. 
"all ranges" is a good name. also need a good short name for a list of cidrs that are not "all ranges"

haim-kermany:
its a boolean on the conn string. if the label is "alla connection", it should return true

ShiriMoran:
> its a boolean on the conn string. if the label is "alla connection", it should return true

ahh. So you don't care if the connection is from/to all subnets/all vsis in a subnet (kitbag question, I know)

haim-kermany:
can we use Kind(), and remove Type()

zivnevo:
From today discussion:
* The `DetailsMap` mechanism used today is going to be hard to continue with.
* From a software engineering perspective, we mix in `drawioOutput.go` abstract and concrete (vendor-specific) implementations. Moreover, in the abstract library we rely on strings that are defined in the concrete library. This is going to bite us in the future.
[...]

PR #177, opened 2023-09-28
remove Details() and DetailsMap() from VPCResourceIntf

adisos:
Removing `Details()` and `DetailsMap()` following the discussion from issue #144 , see:  https://github.com/np-guard/vpc-network-config-analyzer/issues/144#issuecomment-1707983895 

Additional changes followed by this change:
- Removing `arch` part of the `json` output


zivnevo (approved):
I love it when so much code just disappears 😸

Issue #182, opened 2023-10-02
unit test TestSelfLoopCliqueDiffSubnets fails occasionly 


ShiriMoran:
e.g.  https://github.com/np-guard/vpc-network-config-analyzer/actions/runs/6372885986/job/17296113715?pr=181

ShiriMoran:
solved by #183

PR #183, opened 2023-10-02
merging groups treating self loops as don't cares might result in an …

ShiriMoran:
…already existing key. In which case we must not override.

E.g:
a -> b
b -> c
a, b -> d
must result in
a, b -> a, b, d


adisos (approved):

ShiriMoran:
solves #182

PR #185, opened 2023-10-02
Avoid discarding errors in CIDR parsing

elazarg:


elazarg:
I don't think the lint error (function too long) makes sense to address in this PR

adisos:
> I don't think the lint error (function too long) makes sense to address in this PR

You can either split the function or add an instruction to ignore that linter for this function.

adisos (approved):

PR #188, opened 2023-10-05
Make hypercube and interval internal packages

elazarg:
Two semantic changes:
* isSingleNumber becomes public
* IntervalToCidrList becomes private, and accepts an interval instead of a pair


zivnevo:
Packages moved to `models` repo

Issue #157, opened 2023-09-14
json output follow-up items


adisos:
https://github.com/np-guard/vpc-network-config-analyzer/pull/133#pullrequestreview-1626710361

              I have no comment that should block merging this PR.
Please see my questions and suggestions below - maybe for future work...

_Originally posted by @zivnevo in https://github.com/np-guard/vpc-network-config-analyzer/pull/133#pullrequestreview-1626710361_

adisos:
remaining items:

- [ ] data_model.go : temporary solution until we can import the jsonio package from vpc-network-config-synthesis
- [ ] handle comparison of json output in tests

PR #201, opened 2023-10-18
Sort json output

adisos:
issue #157

### DIFF
--- a/pkg/vpcmodel/jsonOutput.go
+++ b/pkg/vpcmodel/jsonOutput.go
@@ -16,7 +16,6 @@ type connLine struct {
 	Dst                EndpointElem       `json:"dst"`
 	Conn               common.ConnDetails `json:"conn"`
 	UnidirectionalConn common.ConnDetails `json:"unidirectional_conn,omitempty"`
-	connStr            string             // connStr used for sorting
 }
 
 func sortConnLines(connLines []connLine) {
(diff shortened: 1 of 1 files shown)


pkg/vpcmodel/jsonOutput.go
zivnevo, line 1: Please fix the error message at the end of the file, which refers to `md` instead to `json`

pkg/vpcmodel/jsonOutput.go
```diff
@@ -15,6 +16,20 @@ type connLine struct {
...
+func sortConnLines(connLines []connLine) {
+	sort.Slice(connLines, func(i, j int) bool {
+		if connLines[i].Src.Name() != connLines[j].Src.Name() {
+			return connLines[i].Src.Name() < connLines[j].Src.Name()
+		}
+		if connLines[i].Dst.Name() != connLines[j].Dst.Name() {
+			return connLines[i].Dst.Name() < connLines[j].Dst.Name()
+		}
+
+		return connLines[i].connStr < connLines[j].connStr
```
zivnevo, line 31: How can we have two lines with the same <src, dst>?
Seems to me like we don't really need `connStr`.

zivnevo (changes requested):

pkg/vpcmodel/jsonOutput.go
adisos: Fixed.

pkg/vpcmodel/jsonOutput.go
adisos: Right, fixed.

zivnevo (approved):
LGTM

PR #203, opened 2023-10-19
163 semantic diff subnets a

ShiriMoran:
Diff between subnets based connections
Missing:
Diff between connections with external addresses
Command line parms
Grouping (different issue)
Enhance connection diff (different issue)


adisos (commented):
Looks good, added a few comments

pkg/vpcmodel/semanticDiffSubnets.go
```diff
@@ -0,0 +1,212 @@
...
+// ToDo: this is done based on names only at the moment. Perhaps take into account other factors such as cidr?
+// ToDo: instead of performing this search each time, use a map created once
+func (c *CloudConfig) getEndpointElemInOtherConfig(other *CloudConfig, ep EndpointElem) EndpointElem {
+	if ep.IsExternal() {
+		for _, node := range other.Nodes {
+			if node.Name() == ep.Name() {
+				res := EndpointElem(node)
+				return res
+			}
+		}
```
adisos, line 84: I don't think that external ip-blocks can be considered as "missing endpoint" in a config. The connectivity result per config should contain connectivity per all ip-blocks ranges.

pkg/vpcmodel/grouping_test.go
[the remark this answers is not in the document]
ShiriMoran: I added a test for subnet analysis on something that is already working as a preliminary stage to the unit test - to make sure the subnet tests is written properly; otherwise it would have been harder to debug both the subnets unit test and the code itself where both most probably contains errors. 
Then I had 3 options:
1. Remove the added grouping unit test 
1. Submit a separate PR for the grouping test
1. Submit as part of the current PR
And I choose the last

pkg/vpcmodel/semanticDiffSubnets.go
```diff
@@ -0,0 +1,212 @@
...
+	subnet1Subtract2 SubnetsDiff
+	subnet2Subtract1 SubnetsDiff
+
+	GroupedSubnet1Minus2 *GroupConnLines
+	GroupedSubnet1Minus1 *GroupConnLines
+}
+
+func (configs ConfigsForDiff) GetSubnetsDiff(grouping bool) (*diffBetweenSubnets, error) {
+	// 1. compute connectivity for each of the subnets
+	subnetsConn1, err := configs.config1.GetSubnetsConnectivity(true, grouping)
```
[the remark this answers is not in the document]
ShiriMoran: Actually the grouping is not relevant when calling GetSubnetsConnectivity , since the grouping of the diffs will be done separately, in any stage. Fixed.

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: The current PR and its unit test does not actually handle external ranges. So let me comment it for now and add todo, and will be handled in the next PR that handles external addresses.

pkg/vpcmodel/diffSubnets_test.go
```diff
@@ -0,0 +1,90 @@
...
+	subnet1Subtract2 := subnetConfigConn1.SubnetConnectivitySubtract(subnetConfigConn2)
+	subnet1Subtract2Str := subnet1Subtract2.EnhancedString(true)
+	fmt.Printf("subnet1Subtract2:\n%v\n", subnet1Subtract2Str)
+	newLines := strings.Count(subnet1Subtract2Str, "\n")
+	// there should be 4 lines in subnet1Subtract2Str
+	require.Equal(t, 4, newLines)
+	require.Contains(t, subnet1Subtract2Str, "-- subnet3 => subnet1 :  missing destination")
+	require.Contains(t, subnet1Subtract2Str, "-- subnet2 => subnet3 :  missing connection")
+	require.Contains(t, subnet1Subtract2Str, "-- subnet0 => subnet1 :  missing source and destination")
+	require.Contains(t, subnet1Subtract2Str, "-- subnet1 => subnet2 :  missing source")
```
adisos, line 84: Extend to include the case of a changed connection

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: fixed

pkg/vpcmodel/diffSubnets_test.go
ShiriMoran: 1. Currently the functionality used to compute the delta between connections is an old one
2. There is no example of unit test with non-trivial connection, and we lack the required functionality for that which needs to be added. So I suggest we defer this to the PR in which we will deal with connections (in the near future).

pkg/vpcmodel/semanticDiffSubnets.go
```diff
@@ -0,0 +1,221 @@
...
+		for dst, conns := range endpointConns {
+			if conns.IsEmpty() {
+				continue
+			}
+			if _, ok := connectivitySubtract[src]; !ok {
+				connectivitySubtract[src] = map[EndpointElem]*connectionDiff{}
+			}
+			diffConnectionWithType := &connectionDiff{nil, NoDiff}
+			srcInOther := subnetConfConnectivity.config.getEndpointElemInOtherConfig(other.config, src)
+			dstInOther := subnetConfConnectivity.config.getEndpointElemInOtherConfig(other.config, dst)
```
adisos, line 126: if `src` or `dst` is an external ip-range, it could be a one that originally was a wider range and then got split for alignment of ip-ranges for both connectivity-maps, so it may be an issue to search for the exact same ip-range in the config nodes..

pkg/vpcmodel/diffSubnets_test.go
adisos: You can create in the `VPCsubnetConnectivity` an edge between two subnets with a connection-set that is not the trivial `NewConnectionSet(true)`, such as in https://github.com/np-guard/vpc-network-config-analyzer/blob/main/pkg/ibmvpc/connectivityAnalysis_test.go#L95-L99

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: Sure, but this part is not implemented yet, as I mentioned above.
This is why it does not have a unit test either, yet.
Please see documentation to the not yet written function `getComparableConnections` (I'm working on it for the next PR)

ShiriMoran: Inserted changed connection unit test

adisos: OK, just thought the current code is a bit misleading, maybe add a comment that this should be changed and the ip-range cannot be necessarily found as-is in the config.

ShiriMoran: I mentioned it above, and the code is not yet connected to the command line and thus can not actually be used. I'm working in the ip-range now. Comment also added.

adisos (approved):

Issue #53, opened 2023-04-27
prepare support for multi-vpc config


adisos:
- [x] model: e.g. vpc should point to its subnets 
   Remaining items: 
(1) [done] keep in VPCConfig an abstract member of type VPCResourceIntf which points to the concrete VPC the   config belongs to, and delete vpcsMap 
(2) [done] support with formats: json
(3) change map from object name to object reference in vpc config to be based on unique ids (see issue #256 )
(4) decide on JSON format for aggregation - list or map (see https://github.com/np-guard/vpc-network-config-analyzer/pull/207#discussion_r1378639922 )
(5) handle drawio output for multiple vpcs , see https://github.com/np-guard/vpc-network-config-analyzer/pull/207#discussion_r1378639410 

- [x] should add a flag to choose which vpc to analyze

see issue #164

Issue #129, opened 2023-08-07
Improve handling of info message about unidriectional connections


adisos:
see comments from:

https://github.com/np-guard/vpc-network-config-analyzer/pull/125#discussion_r1285416364 

few options to consider: 
- the info message should / should not be always added to the output
- the info message should be added once, not from various code functions

current state: info msg is `"connections are stateful unless marked with *"`, and is always added to textual output (txt,md)

- [x] #292
- [x] #294
- [x] Avoid involvement in unit tests expected output ?

Make sure that the info message will be added once, not from various code functions

consider also the comment from another PR : https://github.com/np-guard/vpc-network-config-analyzer/pull/207#discussion_r1378485943 
The report of aggregated VPCs should contain this line only once rather than after every VPC report.

adisos:
- [ ] The report of aggregated VPCs should contain this line only once rather than after every VPC report.
- [ ] The info message should not be always added to the output
- [ ] Avoid involvement in unit tests expected output ?
- [ ] the info message should be added once, not from various code functions

PR #207, opened 2023-10-31
Issue 53 additional updates

adisos:
issue #53 follow-ups:

(1) keep in VPCConfig an abstract member of type VPCResourceIntf which points to the concrete VPC the config belongs to, and delete vpcsMap
(2) support multiple vpcs with format JSON

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -40,7 +40,7 @@ func analysisTypeToUseCase(inArgs *InArgs) vpcmodel.OutputUseCase {
 	return vpcmodel.AllEndpoints
 }
 
-func analysisPerVPCConfig(c *vpcmodel.VPCConfig, inArgs *InArgs) (*vpcmodel.VPCAnalysisOutput, error) {
+func analysisPerVPCConfig(c *vpcmodel.VPCConfig, inArgs *InArgs, outFile string) (*vpcmodel.VPCAnalysisOutput, error) {
 	og, err := vpcmodel.NewOutputGenerator(c,
 		*inArgs.Grouping,
 		analysisTypeToUseCase(inArgs),
@@ -49,8 +49,13 @@ func analysisPerVPCConfig(c *vpcmodel.VPCConfig, inArgs *InArgs) (*vpcmodel.VPCA
 		return nil, err
 	}
 
+	var genOutFile string
+	// currently for drawio output only one vpc level is supported, and not as aggregated output of multiple vpcs
+	if *inArgs.OutputFormat == ARCHDRAWIOFormat || *inArgs.OutputFormat == DRAWIOFormat {
+		genOutFile = outFile
+	}
 	outFormat := getOutputFormat(inArgs)
-	output, err := og.Generate(outFormat, "")
+	output, err := og.Generate(outFormat, genOutFile)
 	if err != nil {
 		return nil, fmt.Errorf("output generation error: %w", err)
 	}
@@ -78,11 +83,16 @@ func _main(cmdlineArgs []string) error {
 	if err != nil {
 		return fmt.Errorf("error generating cloud config from input vpc resources file: %w", err)
 	}
+	outFile := ""
+	if inArgs.OutputFile != nil {
+		outFile = *inArgs.OutputFile
+	}
+
 	outputPerVPC := make([]*vpcmodel.VPCAnalysisOutput, len(vpcConfigs))
 	i := 0
 	var vpcAnalysisOutput *vpcmodel.VPCAnalysisOutput
 	for _, vpcConfig := range vpcConfigs {
-		vpcAnalysisOutput, err = analysisPerVPCConfig(vpcConfig, inArgs)
+		vpcAnalysisOutput, err = analysisPerVPCConfig(vpcConfig, inArgs, outFile)
 		if err != nil {
 			return err
 		}
@@ -90,10 +100,6 @@ func _main(cmdlineArgs []string) error {
 		i++
 	}
 
-	outFile := ""
-	if inArgs.OutputFile != nil {
-		outFile = *inArgs.OutputFile
-	}
 	out, err := vpcmodel.AggregateVPCsOutput(outputPerVPC, getOutputFormat(inArgs), outFile)
 	if err != nil {
 		return err
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -90,11 +90,10 @@ func _main(cmdlineArgs []string) error {
 
 	outputPerVPC := make([]*vpcmodel.VPCAnalysisOutput, len(vpcConfigs))
 	i := 0
-	var vpcAnalysisOutput *vpcmodel.VPCAnalysisOutput
 	for _, vpcConfig := range vpcConfigs {
-		vpcAnalysisOutput, err = analysisPerVPCConfig(vpcConfig, inArgs, outFile)
-		if err != nil {
-			return err
+		vpcAnalysisOutput, err2 := analysisPerVPCConfig(vpcConfig, inArgs, outFile)
+		if err2 != nil {
+			return err2
 		}
 		outputPerVPC[i] = vpcAnalysisOutput
 		i++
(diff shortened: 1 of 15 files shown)


pkg/vpcmodel/output.go
```diff
@@ -64,7 +67,16 @@ func NewOutputGenerator(c *VPCConfig, grouping bool, uc OutputUseCase, archOnly
...
+// VPCAnalysisOutput captures output per VPC
+type VPCAnalysisOutput struct {
+	VPCName    string
+	Output     string
+	jsonStruct interface{}
+	format     OutFormat
+}
+
+// Generate returns VPCAnalysisOutput for its VPC analysis results
+func (o *OutputGenerator) Generate(f OutFormat, outFile string) (*VPCAnalysisOutput, error) {
```
zivnevo, line 79: Is there a reason for `Generate` to still get an `outFile`?
If it's only for testing, I would find other solutions.
Currently, formatters call `writeOutput` or `WriteToFile` (do we really need both?) without any effect.

pkg/ibmvpc/examples/multiple_vpcs_all_vpcssubnetsBased_withPGW.txt
```diff
@@ -0,0 +1,98 @@
...
+ky-testenv-transit-subnet-3 => ky-testenv-edge-subnet-1 : All Connections
+ky-testenv-transit-subnet-3 => ky-testenv-edge-subnet-2 : All Connections
+ky-testenv-transit-subnet-3 => ky-testenv-edge-subnet-3 : All Connections
+ky-testenv-transit-subnet-3 => ky-testenv-private-subnet-1 : protocol: TCP src-ports: 443
+ky-testenv-transit-subnet-3 => ky-testenv-private-subnet-2 : protocol: TCP src-ports: 443
+ky-testenv-transit-subnet-3 => ky-testenv-private-subnet-3 : protocol: TCP src-ports: 443
+ky-testenv-transit-subnet-3 => ky-testenv-transit-subnet-1 : All Connections
+ky-testenv-transit-subnet-3 => ky-testenv-transit-subnet-2 : All Connections
+
+connections are stateful unless marked with *
```
zivnevo, line 79: Would be nice if we can have this comment only once per report rather than after every VPC report.

pkg/ibmvpc/examples/experiments_env_all_vpcs.json
```diff
@@ -0,0 +1,6078 @@
+{
+    "test-vpc1-ky": {
+        "endpoints_connectivity": [
+            {
+                "src": {
+                    "ResourceType": "Public Internet",
```
zivnevo, line 6:
```suggestion
{
    "test-vpc1-ky": {
        "endpoints_connectivity": [
            {
                "src": {
                    "ResourceType": "Public Internet",
```
```suggestion
[
    {
        "vpc_name": "test-vpc1-ky",
        "endpoints_connectivity": [
            {
                "src": {
                    "ResourceType": "Public Internet",
```
Would something like this be better (in case we would like to provide more VPC details in the future)? What do you think?

cmd/analyzer/main.go
```diff
@@ -84,12 +78,27 @@ func _main(cmdlineArgs []string) error {
 	if err != nil {
 		return fmt.Errorf("error generating cloud config from input vpc resources file: %w", err)
 	}
+	outputPerVPC := make([]*vpcmodel.VPCAnalysisOutput, len(vpcConfigs))
+	i := 0
+	var vpcAnalysisOutput *vpcmodel.VPCAnalysisOutput
 	for _, vpcConfig := range vpcConfigs {
-		err = analysisPerVPCConfig(vpcConfig, inArgs)
+		vpcAnalysisOutput, err = analysisPerVPCConfig(vpcConfig, inArgs)
```
zivnevo, line 85:
```suggestion
		vpcAnalysisOutput, err := analysisPerVPCConfig(vpcConfig, inArgs)
```
No need for the var defined outside of the loop

cmd/analyzer/main.go
zivnevo, line 1: It turned out to be quite a bit of code that calls the analysis for each VPC, then aggregates the outputs together.
Consider moving some of this code to a package (preferably in a follow-up PR).

zivnevo (changes requested):

cmd/analyzer/main.go
adisos: I originally wrote as you suggested, but then I got a linter issue ` shadow: declaration of "err" shadows declaration at line ...`

pkg/ibmvpc/examples/experiments_env_all_vpcs.json
adisos: I think it might be better for having the option to add more vpc details, though Karen asked for a map instead of list.. @kyorav  what do you think?

pkg/vpcmodel/output.go
adisos: - Deleted function `writeOutput`
- Actually `Generate` should still get an `outFile` because of the `drawio` format -- the aggregation does not yet support this format.. added a fix to call `Generate` with `outFile` in that case for now.

pkg/ibmvpc/examples/multiple_vpcs_all_vpcssubnetsBased_withPGW.txt
adisos: Will be handled in a related open issue #129

cmd/analyzer/main.go
zivnevo: you can simply rename `err` to `err2` on such cases

cmd/analyzer/main.go
adisos: Opened new issue #212 for this.

cmd/analyzer/main.go
adisos: ok, changed

pkg/vpcmodel/output.go
zivnevo: @haim-kermany, is there a way to have multiple diagrams in a single .drawio file?

pkg/ibmvpc/examples/experiments_env_all_vpcs.json
kyorav: I am not following what exactly is the question. Keeping the info as a map does not preclude adding more keys to the object, including a "vpc_name" key, does it?
Anyway, the argument for making this into a map is extremely thin, if it's better as an array then go for it.

pkg/vpcmodel/textOutput.go
```diff
@@ -29,5 +29,6 @@ func (t *TextOutputFormatter) WriteOutput(c *VPCConfig,
 		out += c.GetConnectivityOutputPerEachSubnetSeparately()
 	}
 	// write output to file and return the output string
-	return writeOutput(out, outFile)
+	outStr, err := WriteToFile(out, outFile)
+	return &VPCAnalysisOutput{Output: outStr, VPCName: c.VPC.Name(), format: Text}, err
```
zivnevo, line 33:
```suggestion
	_, err := WriteToFile(out, outFile)
	return &VPCAnalysisOutput{Output: out, VPCName: c.VPC.Name(), format: Text}, err
```

pkg/vpcmodel/debugOutput.go
```diff
@@ -20,5 +20,6 @@ func (t *DebugOutputFormatter) WriteOutput(c *VPCConfig,
 	case SingleSubnet:
 		out = c.GetConnectivityOutputPerEachSubnetSeparately()
 	}
-	return writeOutput(out, outFile)
+	outStr, err := WriteToFile(out, outFile)
+	return &VPCAnalysisOutput{Output: outStr, VPCName: c.VPC.Name(), format: Debug}, err
```
zivnevo, line 24:
```suggestion
	_, err := WriteToFile(out, outFile)
	return &VPCAnalysisOutput{Output: out, VPCName: c.VPC.Name(), format: Debug}, err
```

zivnevo (changes requested):
Tiny things

zivnevo (approved):

pkg/vpcmodel/output.go
haim-kermany: It can be done

PR #208, opened 2023-10-31
166 single group element

ShiriMoran:


pkg/vpcmodel/grouping.go
```diff
@@ -332,6 +333,36 @@ func (g *GroupConnLines) groupInternalSrcOrDst(srcGrouping, groupVsi bool) {
...
+func (g *GroupConnLines) unifiedGroupedConnLines(oldConnLines []*GroupedConnLine) []*GroupedConnLine {
+	newGroupedLines := []*GroupedConnLine{}
+	// go over all connections; if src/dst is not external then use groupedEndpointsElemsMap
+	for _, groupedConnLine := range oldConnLines {
+		newGroupedConnLine := &GroupedConnLine{g.unifiedGroupedElems(groupedConnLine.Src),
+			g.unifiedGroupedElems(groupedConnLine.Dst),
+			groupedConnLine.Conn}
+		newGroupedLines = append(newGroupedLines, newGroupedConnLine)
+	}
+	return newGroupedLines
```
adisos, line 348:
```suggestion
	newGroupedLines := make([]*GroupedConnLine, len(oldConnLines))
	// go over all connections; if src/dst is not external then use groupedEndpointsElemsMap
	for i, groupedConnLine := range oldConnLines {
		newGroupedLines[i] = &GroupedConnLine{g.unifiedGroupedElems(groupedConnLine.Src),
			g.unifiedGroupedElems(groupedConnLine.Dst),
			groupedConnLine.Conn}
	}
	return newGroupedLines
```

pkg/vpcmodel/grouping.go
```diff
@@ -306,6 +306,7 @@ func getKeyOfGroupConnLines(ep EndpointElem, connection string) string {
 // assuming the  g.groupedLines was already initialized by previous step groupExternalAddresses()
 func (g *GroupConnLines) groupInternalSrcOrDst(srcGrouping, groupVsi bool) {
 	res, groupingSrcOrDst := g.groupLinesByKey(srcGrouping, groupVsi)
+	res = g.unifiedGroupedConnLines(res)
```
adisos, line 309: How is `res` affected by `extendGroupingSelfLoops`? It seems that only `groupLinesByKey` adds lines to `res`,  so why on this stage the func `unifiedGroupedConnLines` works on `res`?

haim-kermany:
the code works as eccpected, tnx

pkg/vpcmodel/grouping.go
ShiriMoran: I'm ok with the change, but in what aspect is it better? @adisos

adisos: More efficient to allocate in advance the array, rather than using `append`

pkg/vpcmodel/grouping.go
ShiriMoran: @adisos good catch. So there are 2 questions to answer here: yours, and why still the fix that seems misplaced solved the issue.
The original problem here is that `extendGroupingSelfLoops` "mess up with"  the basic result of the grouping, modifying both sides and not only the targeted one. This is why we had a problem in the first place. 
The fix I've put in indeed does not cover the problem, as you say. The reason it did seem to cover the problem is that is fixed the result `res` of the previous stage and for the case inline it was suffice. Clearly this is not a proper solution. I'll fix.

ShiriMoran: fixed @haim-kermany please verify it is working as expected (by my testing it is)

adisos (approved):

haim-kermany (approved):

PR #223, opened 2023-11-09
163 output txt fmt

ShiriMoran:


pkg/vpcmodel/semanticDiffSubnets.go
```diff
@@ -126,25 +126,23 @@ func (subnetConfConnectivity *SubnetConfigConnectivity) subtract(other *SubnetCo
...
 			}
+			connDiff := &connectionDiff{conns, nil, MissingConnection}
 			if srcInOther != nil && dstInOther != nil {
 				if otherSrc, ok := other.subnetConnectivity[srcInOther]; ok {
-					if otherSrcDst, ok := otherSrc[dstInOther]; ok {
+					if otherConn, ok := otherSrc[dstInOther]; ok {
 						// ToDo: https://github.com/np-guard/vpc-network-config-analyzer/issues/199
-						subtractConn := conns.Subtract(otherSrcDst)
+						subtractConn := conns.Subtract(otherConn)
 						if subtractConn.IsEmpty() {
```
adisos, line 135:
```suggestion
						if conns.Equals(otherConn) {
```

pkg/vpcmodel/semanticDiffSubnets.go
adisos, line 1: Does it make sense to keep the type `DiffBetweenSubnets` as two parts `subnet1Subtract2` and `subnet2Subtract1` ?
or to keep the function `subtract` with its name? 
As we do not really subtract?

pkg/vpcmodel/semanticDiffSubnets.go
```diff
@@ -203,20 +200,39 @@ func (subnetDiff *SubnetsDiff) EnhancedString(thisMinusOther bool) string {
...
+func connStr(conn *common.ConnectionSet) string {
+	if conn == nil {
+		return "No connection"
+	}
+	return conn.EnhancedString()
+}
+
+func diffAndWorkLoadDisc(diff DiffType, src, dst VPCResourceIntf, thisMinusOther bool) (diffDisc, workLoad string) {
+	const (
+		WorkloadDiffInfo = ", workloads-diff-info: workload"
```
adisos, line 213: There is no term `workload` here.. for the subnets analysis it is `subnet`, and for endpoints it is `network interface`

adisos, line 213: why exporting these const values?
There are other const definitions in this file that should not be exported

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: There are actually two questions here:
1. Do we Subtract? before we did. Perhaps now we are not exactly subtracting, we can rename
2. Does it make sense to have an operand between the cfgs that is not symmetric and that has to be operated in both ways to get the diff. I believe the answer here is positive. This puts some burden on the printing function, but all at all its better than having a messy function that will replace the two Subtracts and keeping it in one part. I believe it is much smoother this way.

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: I asked you whether this is the term to use, and I thought you said yes. Clearly this was a misunderstanding - I'll fix. It certainly makes more sense this way.

pkg/ibmvpc/examples/acl_testing5subnetsDiff.txt
```diff
@@ -1,9 +1,9 @@
 Analysis for diff between VPC test-vpc-ky1 and VPC test-vpc-ky2
--- sub1-2-ky => sub1-1-ky : missing connection 
--- sub1-3-ky => sub1-1-ky : missing connection 
+diff-type: added source: sub1-2-ky destination: sub1-1-ky dir1: protocol: TCP dir2: No connection
```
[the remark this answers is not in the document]
ShiriMoran: OK

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: No, this change is no good.
The reason is the discussion in the next item - the operand I calculate is not symmetric, as is`subtractConn.IsEmpty()`; however,  `conns.Equals(otherConn` is symmetric.

pkg/ibmvpc/examples/acl_testing5subnetsDiff.txt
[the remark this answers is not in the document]
ShiriMoran: > so an item indicating if one of the subnets was added/removed, appears only if relevant? would it help to add this item with empty value in case the subnet was not added/removed, for consistency of lines?

You mean write "subnets-diff-info:" and keep it empty?

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: done and enhanced to avoid dup lines in final presentation

pkg/vpcmodel/semanticDiffSubnets.go
adisos: 1. Let's rename then
2. Not necessarily need a function that replaces the two subtracts and keep them in one part, could just merge the results from both subtracts into a single `SubnetsDiff`.

pkg/ibmvpc/examples/acl_testing5subnetsDiff.txt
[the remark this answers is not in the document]
adisos: Yes, I think adding this part `subnets-diff-info` will let the output lines be more consistent, and also match the other planned formats (e.g. `md` format as a table where this is another column that either has an empty cell or contain value..).

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: 1. ok
2. Correct, but I don't see why it would be better. If its not too big of a deal for you, I prefer to keep it the way it is.

pkg/vpcmodel/semanticDiffSubnets.go
```diff
@@ -203,20 +202,39 @@ func (subnetDiff *SubnetsDiff) EnhancedString(thisMinusOther bool) string {
...
-func diffDescription(diff DiffType) string {
+// prints connection for func (subnetDiff *SubnetsDiff) EnhancedString(..) where the connection could be empty
+func connStr(conn *common.ConnectionSet) string {
+	if conn == nil {
+		return "No connection"
+	}
+	return conn.EnhancedString()
+}
+
+func diffAndEndpointsDisc(diff DiffType, src, dst VPCResourceIntf, thisMinusOther bool) (diffDisc, workLoad string) {
```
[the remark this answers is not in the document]
ShiriMoran: oops, missed this one. Will do.

pkg/vpcmodel/semanticDiffSubnets.go
adisos: ok

pkg/ibmvpc/examples/acl_testing5subnetsDiff.txt
```diff
@@ -1,9 +1,9 @@
 Analysis for diff between VPC test-vpc-ky1 and VPC test-vpc-ky2
--- sub1-2-ky => sub1-1-ky : missing connection 
--- sub1-3-ky => sub1-1-ky : missing connection 
+diff-type: added, source: sub1-2-ky, destination: sub1-1-ky, config1: protocol: TCP, config2: No connection
+diff-type: added, source: sub1-3-ky, destination: sub1-1-ky, config1: protocol: TCP, config2: No connection
+diff-type: changed, source: sub2-1-ky, destination: Public Internet [8.8.8.8/32], config1: protocol: UDP dst-ports: 53, config2: protocol: UDP dst-ports: 43,53
```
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/semanticDiffSubnets.go
ShiriMoran: done

pkg/ibmvpc/examples/acl_testing5subnetsDiff.txt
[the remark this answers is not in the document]
ShiriMoran: done, though I don't like the way it looks now

pkg/ibmvpc/examples/acl_testing5subnetsDiff.txt
```diff
@@ -1,9 +1,10 @@
...
+diff-type: changed, source: sub1-1-ky, destination: sub1-2-ky, config1: protocol: TCP, config2: protocol: TCP *, subnets-diff-info: 
+diff-type: changed, source: sub1-1-ky, destination: sub1-3-ky, config1: protocol: TCP, config2: protocol: TCP *, subnets-diff-info: 
+diff-type: changed, source: sub2-1-ky, destination: Public Internet [8.8.8.8/32], config1: protocol: UDP dst-ports: 53, config2: protocol: UDP dst-ports: 43,53, subnets-diff-info: 
+diff-type: removed, source: sub2-1-ky, destination: Public Internet [8.8.8.0/29], config1: No connection, config2: protocol: UDP dst-ports: 53 *, subnets-diff-info: 
+diff-type: removed, source: sub2-1-ky, destination: Public Internet [8.8.8.10/31], config1: No connection, config2: protocol: UDP dst-ports: 53 *, subnets-diff-info: 
+diff-type: removed, source: sub2-1-ky, destination: Public Internet [8.8.8.12/30], config1: No connection, config2: protocol: UDP dst-ports: 53 *, subnets-diff-info: 
+diff-type: removed, source: sub2-1-ky, destination: Public Internet [8.8.8.9/32], config1: No connection, config2: protocol: UDP dst-ports: 53 *, subnets-diff-info:
```
adisos, line 10: It seems the terms `added` and `removed` should be used the other way around: for example, line 2, the connection exists in config1 but does not exist in config2, so should be considered as a "removed connection".

ShiriMoran: ok, done

adisos (approved):

PR #224, opened 2023-11-09
ignore fip without target, instead of returning an error

adisos:


pkg/ibmvpc/parser.go
```diff
@@ -516,6 +516,10 @@ func getPgwConfig(
 	return nil
 }
 
+func ignoreFIPWarning(fipName string) string {
+	return "warning: ignoring floatingIP " + fipName
```
zivnevo, line 520:
```suggestion
	return "warning: ignoring floating IP " + fipName + ": unsupported floating IP target."
```
We should provide the user with some clues why this happens.

zivnevo (approved):

pkg/ibmvpc/parser.go
adisos: added more details

PR #229, opened 2023-11-16
209 semantic diff grouping

ShiriMoran:


ShiriMoran:
replaced by https://github.com/np-guard/vpc-network-config-analyzer/pull/230

PR #230, opened 2023-11-19
209 grouping semantic diff new

ShiriMoran:
replaces #229

### DIFF
--- a/pkg/vpcmodel/grouping.go
+++ b/pkg/vpcmodel/grouping.go
@@ -11,6 +11,11 @@ import (
 const commaSeparator = ","
 
 // for each line here can group list of external nodes to cidrs list as of one element
+// TODO: consider wrapping []Node from the map values in the type groupingConnections ,
+//
+//	as a type that includes more than that: []Node + the actual ConnectionSet object +
+//	actual diff-related objects (conn1 object, conn2 object, etc).
+//	It should include all the common relevant objects, aside from the relevant list of nodes that are grouped.
 type groupingConnections map[EndpointElem]map[string][]Node
 
 func (g *groupingConnections) getGroupedConnLines(groupedConnLines *GroupConnLines,
@@ -473,6 +478,9 @@ func (g *groupedExternalNodes) String() string {
 //  2. connection of config1
 //  3. connection of config2
 //  4. info regarding missing endpoints: e.g. vsi0 removed
+//
+// this encoding prevents the need to change all the grouping datastuctures
+// along the pipe: srcToDst and dstToSrc in addition to GroupedConnLine
 func connDiffEncode(src, dst VPCResourceIntf, connDiff *connectionDiff,
 	thisMinusOther bool) string {
 	conn1Str, conn2Str := conn1And2Str(connDiff, thisMinusOther)
--- a/pkg/vpcmodel/grouping.go
+++ b/pkg/vpcmodel/grouping.go
@@ -55,12 +55,23 @@ func newGroupConnLinesSubnetConnectivity(c *VPCConfig, s *VPCsubnetConnectivity,
 	return res
 }
 
+func newGroupConnLinesDiff(d *diffBetweenCfgs) *GroupConnLines {
+	res := &GroupConnLines{d: d,
+		srcToDst:                 newGroupingConnections(),
+		dstToSrc:                 newGroupingConnections(),
+		groupedEndpointsElemsMap: make(map[string]*groupedEndpointsElems),
+		groupedExternalNodesMap:  make(map[string]*groupedExternalNodes)}
+	res.computeGroupingForDiff()
+	return res
+}
+
 // GroupConnLines used both for VPCConnectivity and for VPCsubnetConnectivity, one at a time. The other must be nil
 // todo: define abstraction above both?
 type GroupConnLines struct {
 	c        *VPCConfig
 	v        *VPCConnectivity
 	s        *VPCsubnetConnectivity
+	d        *diffBetweenCfgs
 	srcToDst *groupingConnections
 	dstToSrc *groupingConnections
 	// a map to groupedEndpointsElems used by GroupedConnLine from a unified key of such elements
@@ -262,6 +273,42 @@ func (g *GroupConnLines) groupExternalAddressesForSubnets() {
 	g.GroupedLines = res
 }
 
+func (g *GroupConnLines) groupExternalAddressesForDiff(thisMinusOther bool) {
+	// group public internet ranges
+	// initialize data structures
+	g.srcToDst = newGroupingConnections()
+	g.dstToSrc = newGroupingConnections()
+	var res []*GroupedConnLine
+	var connRemovedChanged connectivityDiff
+	if thisMinusOther {
+		connRemovedChanged = g.d.cfg1ConnRemovedFrom2
+	} else {
+		connRemovedChanged = g.d.cfg2ConnRemovedFrom1
+	}
+	for src, endpointConnDiff := range connRemovedChanged {
+		for dst, connDiff := range endpointConnDiff {
+			connDiffString := EnhancedConnDiffDecode(src, dst, connDiff, g.d.diffAnalysis, thisMinusOther)
+			srcNode, srcIsNode := src.(Node)
+			dstNose, dstIsNode := dst.(Node)
+			srcIsExternal := srcIsNode && src.IsExternal()
+			dstIsExternal := dstIsNode && dst.IsExternal()
+			switch {
+			case dstIsExternal:
+				g.srcToDst.addPublicConnectivity(src, connDiffString, dstNose)
+			case srcIsExternal:
+				g.dstToSrc.addPublicConnectivity(dst, connDiffString, srcNode)
+			default:
+				res = append(res, &GroupedConnLine{src, dst, connDiffString})
+			}
+		}
+	}
+
+	// add to res lines from  srcToDst and DstToSrc groupings
+	res = append(res, g.srcToDst.getGroupedConnLines(g, true)...)
+	res = append(res, g.dstToSrc.getGroupedConnLines(g, false)...)
+	g.GroupedLines = append(g.GroupedLines, res...)
+}
+
 // aux func, returns true iff the EndpointElem is Node if grouping vsis or NodeSet if grouping subnets
 func isInternalOfRequiredType(ep EndpointElem, groupVsi bool) bool {
 	if groupVsi { // groups vsis Nodes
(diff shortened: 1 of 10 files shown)


pkg/vpcmodel/semanticDiff.go
```diff
@@ -271,6 +259,58 @@ func (connDiff *connectivityDiff) string(diffAnalysis diffAnalysisType, thisMinu
...
+		conn2Str = connStr(connDiff.conn1)
+	}
+	return conn1Str, conn2Str
+}
+
+// connDiffEncode decode the above string
+func connDiffEncode(src, dst EndpointElem, decoded string) string {
+	encoded := strings.Split(decoded, semicolon)
+	printDiff := fmt.Sprintf("%s, source: %s, destination: %s, %s\n", encoded[0], src.Name(), dst.Name(), encoded[1])
+	return printDiff
```
adisos, line 301: what is `decoded`? what is `encoded[0]` , `encoded[1]` ?
can you avoid relying on strings formats and splitting strings? it is usually error-prone and more difficult to maintain the code, as printed formats may change..

pkg/vpcmodel/grouping.go
```diff
@@ -262,6 +273,42 @@ func (g *GroupConnLines) groupExternalAddressesForSubnets() {
 	g.GroupedLines = res
 }
 
+func (g *GroupConnLines) groupExternalAddressesForDiff(thisMinusOther bool) {
```
adisos, line 276: There are now 3 functions that are very similar: `groupExternalAddressesForDiff` , `groupExternalAddresses`, `groupExternalAddressesForSubnets` -- all are about grouping external ip-blocks.. is it possible to generalize and have only one function?

pkg/vpcmodel/semanticDiff.go
ShiriMoran: 1. We rely on string formats in other cases in grouping as well, by using `func (conn *ConnectionSet) EnhancedString() string`. Keeping the format of the grouping as is is beneficial, and I thing its better to keep it as is. Namely, to encode all info other than `src` and `dst` into a string for grouping. 
2. However, I agree that the decoding should be done more carefully and specifically it should be reused, e.g. by other outputs format that I begun writing code for. So I will change this function so it returns the different semantic components instead of a single string. This way we will have a single encoding func and a single decoding func. This is not too error prone, I believe
3. It also seems that I switched the decode and encode terms. Will fix.

pkg/vpcmodel/grouping.go
ShiriMoran: These function range over different content and do different operations on them. E.g., for `subnets` `src` can not be a public internet. Creating a unified datastructure for which we can have a single function that ranges over it will result with more code (and more functions) which is not readable, I'm afraid.
However, I will try to extract identical (or almost identical) code segments and reuse them.

ShiriMoran: done

pkg/vpcmodel/semanticDiff.go
ShiriMoran: done

pkg/vpcmodel/grouping.go
ShiriMoran: Merged  2 of the 3; extracted reused code for all 3

pkg/vpcmodel/grouping.go
```diff
@@ -214,52 +228,83 @@ func subnetGrouping(groupedConnLines *GroupConnLines,
...
-				res = append(res, &GroupedConnLine{src, dst, connString})
-			}
-		}
+func (g *GroupConnLines) addLineToExternalGrouping(res *[]*GroupedConnLine, emptyConn bool,
+	src, dst VPCResourceIntf, connEnhanced string) error {
+	if emptyConn {
+		return nil
+	}
+	srcNode, srcIsNode := src.(Node)
+	dstNose, dstIsNode := dst.(Node)
```
adisos, line 284:
```suggestion
	dstNode, dstIsNode := dst.(Node)
```

pkg/vpcmodel/semanticDiff.go
```diff
@@ -271,6 +246,59 @@ func (connDiff *connectivityDiff) string(diffAnalysis diffAnalysisType, thisMinu
...
+	} else {
+		conn1Str = connStr(connDiff.conn2)
+		conn2Str = connStr(connDiff.conn1)
+	}
+	return conn1Str, conn2Str
+}
+
+// printDiffLine print one diff line, using the above string as input
+func printDiffLine(diffAnalysis diffAnalysisType, src, dst EndpointElem, decoded string) string {
+	decodedDetails := strings.Split(decoded, semicolon)
```
adisos, line 284: Instead of encoding/decoding strings, can you extend the type `GroupedConnLine` to contain the relevant objects?

pkg/vpcmodel/grouping.go
ShiriMoran: done - not accepting the suggestion since it does not include all occurrences

pkg/vpcmodel/semanticDiff.go
```diff
@@ -263,18 +263,16 @@ func (connDiff *connectivityDiff) string(diffAnalysis diffAnalysisType, thisMinu
...
 func connDiffEncode(src, dst VPCResourceIntf, connDiff *connectionDiff,
-	diffAnalysis diffAnalysisType, thisMinusOther bool) string {
+	thisMinusOther bool) string {
 	conn1Str, conn2Str := conn1And2Str(connDiff, thisMinusOther)
 	diffType, endpointsDiff := diffAndEndpointsDisc(connDiff.diff, src, dst, thisMinusOther)
-	diffInfo := diffInfoStr(diffAnalysis)
-	diffTypeStr := fmt.Sprintf("%v %s", diffTypeStr, diffType)
-	connDiffStr := fmt.Sprintf(configsStr, conn1Str, conn2Str, diffInfo, endpointsDiff)
-	return diffTypeStr + semicolon + connDiffStr
+	return diffType + semicolon + conn1Str + semicolon + conn2Str + semicolon + endpointsDiff
```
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/grouping.go
```diff
@@ -362,20 +407,25 @@ func (g *GroupConnLines) unifiedGroupedElems(srcOrDst EndpointElem) EndpointElem
 	return unifiedGroupedEE
 }
 
-func (g *GroupConnLines) computeGrouping(grouping bool) {
-	g.groupExternalAddresses()
+func (g *GroupConnLines) computeGrouping(vsi, grouping bool) (err error) {
```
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/grouping.go
```diff
@@ -214,52 +228,83 @@ func subnetGrouping(groupedConnLines *GroupConnLines,
...
+		}
+	}
+	g.appendGrouped(&res)
+	return nil
+}
+
+func (g *GroupConnLines) groupExternalAddressesForDiff(thisMinusOther bool) error {
+	// initialize data structures
+	g.srcToDst = newGroupingConnections()
+	g.dstToSrc = newGroupingConnections()
```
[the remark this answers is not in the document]
ShiriMoran: Not for the 2nd call of `groupExternalAddressesForDiff`
Added explanation

pkg/vpcmodel/grouping.go
```diff
@@ -214,52 +228,83 @@ func subnetGrouping(groupedConnLines *GroupConnLines,
...
+	var res []*GroupedConnLine
+	var connRemovedChanged connectivityDiff
+	if thisMinusOther {
+		connRemovedChanged = g.d.cfg1ConnRemovedFrom2
+	} else {
+		connRemovedChanged = g.d.cfg2ConnRemovedFrom1
+	}
+	for src, endpointConnDiff := range connRemovedChanged {
+		for dst, connDiff := range endpointConnDiff {
+			connDiffString := connDiffEncode(src, dst, connDiff, thisMinusOther)
```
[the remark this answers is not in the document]
ShiriMoran: ok

pkg/vpcmodel/grouping.go
```diff
@@ -214,52 +228,83 @@ func subnetGrouping(groupedConnLines *GroupConnLines,
...
-			connString := conns.EnhancedString()
-			if dstNode, ok := dst.(Node); ok && dstNode.IsPublicInternet() {
-				g.srcToDst.addPublicConnectivity(src, connString, dstNode)
-			} else { // since pgw enable only egress src can not be public internet, the above is the only option of public internet
-				// not an external connection in source or destination - nothing to group, just append
-				res = append(res, &GroupedConnLine{src, dst, connString})
-			}
-		}
+func (g *GroupConnLines) addLineToExternalGrouping(res *[]*GroupedConnLine, emptyConn bool,
+	src, dst VPCResourceIntf, connEnhanced string) error {
```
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/semanticDiff.go
ShiriMoran: I think the current solution is better.
The alternative is to extend `GroupedConnLine`  as you suggest and to change the actual grouping, adding the new objects to the indexing when the grouping is done for "diffs". Resulting in more code which will not be easier to maintain. 
Grouping is done after all the relevant computation, and we need these objects only for grouping indexes and for printing. Which is why storing it as string suffices for our needs and simplifies the code and the `GroupedConnLine`  data structure.

adisos: why extending `GroupedConnLine` will change the actual grouping? other fields can be added without changing the indexing or the current grouping mechanism.

ShiriMoran: `addPublicConnectivity` is based on the current encoding. Do you suggest changing that?
Do you want to take it offline?

pkg/vpcmodel/grouping.go
adisos, line 1: consider adding a comment above the lines
```
// for each line here can group list of external nodes to cidrs list as of one element
type groupingConnections map[EndpointElem]map[string][]Node
```
as follows:
```
// TODO: consider wrapping []Node from the map values in the type groupingConnections , as a type that includes more than that: []Node + the actual ConnectionSet object + actual diff-related objects (conn1 object, conn2 object, etc). 
It should include all the common relevant objects, aside from the relevant list of nodes that are grouped. 
```

pkg/vpcmodel/semanticDiff.go
ShiriMoran: done

adisos (approved):

PR #236, opened 2023-11-23
201 semantic diff outputs

ShiriMoran:


pkg/ibmvpc/examples/acl_testing5subnetsDiff.json
```diff
@@ -0,0 +1,212 @@
...
+                {
+                    "protocol": "TCP"
+                }
+            ],
+            "conn2": [
+                {
+                    "protocol": "TCP"
+                }
+            ]
+        },
```
adisos, line 27: why is this `connection changed` if both `conn1` and `conn2` are the same?

pkg/ibmvpc/examples/acl_testing5subnetsDiff.md
```diff
@@ -0,0 +1,12 @@
+# Diff between VPC test-vpc-ky1 and VPC test-vpc-ky2
+## Subnets diff report
+| type | src |  dst | conn1 | conn2 | diff-info |
```
adisos, line 3: the column `diff-info` should be `subnets-diff-info` or `vsis-diff-info` as in the `txt ` format

pkg/vpcmodel/jsonOutput.go
```diff
@@ -108,3 +124,75 @@ func getConnLinesForSubnetsConnectivity(conn *VPCsubnetConnectivity) []connLine
...
+	case missingSrcEP:
+		return "source added"
+	case missingDstEP:
+		return "destination added"
+	case missingSrcDstEP:
+		return "source and destination added"
+	case missingConnection:
+		return "connection added"
+	case changedConnection:
+		return connectionChanged
```
adisos, line 195: define hard-coded strings as `const` values.. cannot be reused with const strings from other formats?

pkg/vpcmodel/jsonOutput.go
```diff
@@ -38,6 +46,14 @@ type connLine struct {
...
 }
 
+type diffLine struct {
+	Diff  string             `json:"diff-type"`
+	Src   EndpointElem       `json:"src"`
+	Dst   EndpointElem       `json:"dst"`
+	Conn1 common.ConnDetails `json:"conn1"`
+	Conn2 common.ConnDetails `json:"conn2"`
+}
+
```
adisos, line 56: missing `diff-info` in case of added/removed subnet/vsi?

adisos (changes requested):

pkg/ibmvpc/examples/acl_testing5subnetsDiff.json
ShiriMoran: Indeed, in my tests I somehow missed it. 
The changed segment is the statefullness.  Why isn't `ConnToJSONRep` considering this aspect? Perhaps we should have this fix in another PR (and delay the merge of README until that PR is ready)

pkg/vpcmodel/output.go
```diff
@@ -150,9 +153,24 @@ func AggregateVPCsOutput(outputList []*VPCAnalysisOutput, f OutFormat, outFile s
...
 		for _, o := range outputList {
-			all[o.VPCName] = o.jsonStruct
+			all[o.VPC1Name] = o.jsonStruct
 		}
 		res, err = writeJSON(all, outFile)
 	}
 	return res, err
 }
+
+func FinalizeDiffOutput(output *SingleAnalysisOutput, f OutFormat, outFile string) (string, error) {
```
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/debugOutput.go
```diff
@@ -20,7 +20,13 @@ func (t *DebugOutputFormatter) WriteOutput(c1, c2 *VPCConfig,
 		out = subnetsConn.String()
 	case SingleSubnet:
 		out = c1.GetConnectivityOutputPerEachSubnetSeparately()
+	case SubnetsDiff, EndpointsDiff:
```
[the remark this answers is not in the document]
ShiriMoran: What do you mean "update headerOfAnalyzedVPC ?

Indeed, excluding AllEndpoints debug is identical to text. Wouldn't it be better to say DEBUG is not supported in these modes it is actually identical to text?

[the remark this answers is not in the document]
adisos: Currently the call to `headerOfAnalyzedVPC ` here is with empty string for vpc2 name, although this is not the case for `diff`. 
I'm ok with supporting `debug` output only for the case where it is different from `text`.

pkg/ibmvpc/examples/acl_testing5subnetsDiff.md
ShiriMoran: done

pkg/vpcmodel/jsonOutput.go
ShiriMoran: done

pkg/vpcmodel/debugOutput.go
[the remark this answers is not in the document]
ShiriMoran: both fixes inserted

pkg/vpcmodel/jsonOutput.go
ShiriMoran: No. For json mode the exact diff type is described - e.g. "source removed" or "source and destination added" or "connection removed". I figured that since this is machine readable, having the info more concise makes more sense.

pkg/vpcmodel/debugOutput.go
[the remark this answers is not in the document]
adisos: is it intentional that `headerOfAnalyzedVPC` gets empty string for vpc2 name now? (because diff not supported)?
if so, should not consider in `SingleAnalysisOutput` the field `VPC2Name`.
for the unsupported usecases I would print a message that the output is given in text format. (currently empty output is returned).

pkg/vpcmodel/mdOutput.go
```diff
@@ -11,35 +11,60 @@ type MDoutputFormatter struct {
...
+		} else {
+			mdTitle = mdSubnetsDiffTitle
+			mdHeader = mdSubnetsDiffHeader
+		}
+		lines := []string{mdTitle, mdHeader}
+		connLines := m.getGroupedDiffOutput(cfgsDiff)
+		sort.Strings(connLines)
+		lines = append(lines, connLines...)
+		out += strings.Join(lines, "\n")
+		out += asteriskDetails
```
adisos, line 59: some common code with `AllEndpoints` can be moved to a function, to avoid duplication.

pkg/vpcmodel/debugOutput.go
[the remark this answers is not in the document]
ShiriMoran: 1. Yes, it gets empty string for DEBUG since diff is not supported for DEBUG. Diff is supported for other formats, hence the field` VPC2Name` in `SingleAnalysisOutput`
2. I changed the parser so that in `DEBUG` only all_endpoints is possible. So no need to print anything in these cases

pkg/vpcmodel/mdOutput.go
```diff
@@ -50,6 +75,16 @@ func (m *MDoutputFormatter) getGroupedOutput(conn *VPCConnectivity) []string {
...
 
+func (m *MDoutputFormatter) getGroupedDiffOutput(diff *diffBetweenCfgs) []string {
+	lines := make([]string, len(diff.groupedLines))
+	for i, line := range diff.groupedLines {
+		decodedDetails := strings.Split(line.Conn, semicolon)
+		lines[i] = fmt.Sprintf("| %s | %s | %s | %s | %s | %s |", decodedDetails[0], line.Src.Name(),
+			line.Dst.Name(), decodedDetails[1], decodedDetails[2], decodedDetails[3])
+	}
+	return lines
+}
```
adisos, line 86: - missing documentation of expected format (an example in a comment could help), for which decoding relies on
- instead of decoding and relying on string formats, can use the added objects to `groupedLines`?

pkg/vpcmodel/debugOutput.go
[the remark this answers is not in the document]
adisos: ok, as for (1) -- I meant that in the function `func (t *DebugOutputFormatter) WriteOutput` , there is no need to compute `v2Name`  and assign it within `SingleAnalysisOutput.VPC2Name`,  since we assume this function is not relevant for `diff` usecases.

pkg/ibmvpc/examples/acl_testing5subnetsDiff.json
adisos: in the link for this issue, seems it was opened in the wrong repo (`vpc-network-config-synthesis ` instead of this repo)

pkg/vpcmodel/jsonOutput.go
```diff
@@ -108,3 +122,85 @@ func getConnLinesForSubnetsConnectivity(conn *VPCsubnetConnectivity) []connLine
...
+	SemanticDiff []diffLine `json:"semantic_diff"`
+}
+
+func getDiffLines(configsDiff *diffBetweenCfgs) []diffLine {
+	diffLines := getDirectionalDiffLines(configsDiff.cfg1ConnRemovedFrom2, true)
+	diffLines = append(diffLines, getDirectionalDiffLines(configsDiff.cfg2ConnRemovedFrom1, false)...)
+	return diffLines
+}
+
+func getDirectionalDiffLines(connectDiff connectivityDiff, thisMinusTheOther bool) []diffLine {
```
adisos, line 136: after the merge from `main` branch -- `thisMinusTheOther` is part of `connectionDiff` object, so there is no need to send it as an argument to this function.

pkg/vpcmodel/debugOutput.go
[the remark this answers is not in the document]
ShiriMoran: right. fixed

pkg/ibmvpc/examples/acl_testing5subnetsDiff.json
ShiriMoran: https://github.com/np-guard/vpc-network-config-analyzer/issues/240

pkg/vpcmodel/mdOutput.go
ShiriMoran: done

pkg/vpcmodel/mdOutput.go
ShiriMoran: already changed in merge with latest main

pkg/vpcmodel/jsonOutput.go
ShiriMoran: right

pkg/vpcmodel/mdOutput.go
```diff
@@ -53,10 +53,7 @@ func (m *MDoutputFormatter) WriteOutput(c1, c2 *VPCConfig,
 		}
 		lines := []string{mdTitle, mdHeader}
 		connLines := m.getGroupedDiffOutput(cfgsDiff)
-		sort.Strings(connLines)
-		lines = append(lines, connLines...)
-		out += strings.Join(lines, "\n")
-		out += asteriskDetails
+		out += linesToOutput(connLines, lines)
```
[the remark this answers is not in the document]
ShiriMoran: fixed

adisos (approved):

Issue #240, opened 2023-11-28
json presentation lacks statefull info


ShiriMoran:
https://github.com/np-guard/vpc-network-config-analyzer/pull/236#discussion_r1406026314

As reflected in `ConnToJSONRep`
dependency: #232

ShiriMoran:
solved by #241

PR #241, opened 2023-11-28
add stateful Info to json format

ShiriMoran:


ShiriMoran:
solves #240

adisos:
Should decide about whether to be consistent with already existing mechanism to compute unidirectional connections subset for endpoints connectivity analysis in the json format. 
Should also decide about whether data_model changes are required.
For now, closing this PR.

PR #243, opened 2023-11-29
228 diff readme

ShiriMoran:


ShiriMoran:
replaced by https://github.com/np-guard/vpc-network-config-analyzer/pull/244

PR #258, opened 2023-12-05
Fix iks node parse isue

adisos:


pkg/ibmvpc/parser.go
```diff
@@ -811,13 +811,13 @@ func getSubnetByCidr(m map[string]*Subnet, cidr string) (*Subnet, error) {
 func getIKSnodesConfig(res map[string]*vpcmodel.VPCConfig,
 	subnetNameToSubnet map[string]*Subnet,
 	rc *ResourcesContainer,
-	skipByVPC func(string) bool) error {
+	skipByVPC func(string) bool) {
 	for _, iksNode := range rc.iksNodes {
 		subnet, err := getSubnetByCidr(subnetNameToSubnet, iksNode.Cidr)
 		if err != nil {
-			return err
+			fmt.Printf("warning: ignoring iksNode with ID %s\n", iksNode.ID)
```
kyorav, line 818:
```suggestion
			fmt.Printf("warning: ignoring iksNode with ID %s (not in VPCs being analyzed)\n", iksNode.ID)
```
Should we add the reason for ignoring?

adisos: at this place we are not sure that the reason is "not in the VPC being analyzed", only that the subnet CIDR referenced by the iks node is not present in the config object as a configured subnet.
How about "could not find subnet with the iksNode's cidr" ?

kyorav: sounds good. capitalize CIDR.

zivnevo (approved):

Issue #42, opened 2023-04-16
Add explainability to the connectivity analysis output


adisos:
Add as a new command line query , input: src / dst (endpoint/subnet/)  , optionally protocol and port to explore 
output: explain which resources (SG/NACL) are relevant for connectivity blocked/allowed , and which rules in the resource.

if blocked: which NACLs is blocking (and if there is deny rule / deny by default) , and which SGs can allow (if exist)
if allowed: which NACLs + SGs allow (and which rules)

phase2: gateways / LBs / PGW (+ ip-address as src/dst )  / TGW

ShiriMoran:
Stage 1: 
input:
src (dst) which is a network interface 
dst (src) which is an external IP
and assuming a single sg that governs the traffic from the given network interface
produce the following output:
1. Is the connection enabled or blocked
2. What are the rules in the sg enabling/blocking the connection or is it just by default with no specific rules addressing this connection

This may be considered as some type of a POC

adisos:
Few comments about proposed output:
- (1) Is the connection enabled or blocked -- if enabled, I think it will also help to add the connection properties.
- (2) default setting - how do we know if a SG is the default one or a one that overrides the default SG? in the input config object contains the configured resources without such indication. (there is no situation where the config object has no SG and the analyzer assumes then a "default" resource to use).

ShiriMoran:
https://github.com/np-guard/vpc-network-config-analyzer/issues/42#issuecomment-1853471622
@adisos I assume you refer to the output below.
1) currently only `SG` are handles, and connections are only enablers. Makes sense, I'll add it.
2) Agrees, but trying to have small `PR`s lets defer it to the next PR. The relevant check that we will need to add here is in a way independent in what we have already done.
```
There is a connection between vsi2-ky[10.240.20.4] and vsi3b-ky[10.240.30.4].
Egress Rules:
~~~~~~~~~~~~~
Security Groups Rules
---------------------
enabling rules from sg2-ky:
        index: 5, direction: outbound, protocol: all, cidr: 10.240.30.0/24
        index: 6, direction: outbound,  conns: protocol: tcp,  dstPorts: 1-65535, cidr: 10.240.20.4/32,10.240.30.4/32

Ingress Rules:
~~~~~~~~~~~~~~
Security Groups Rules
---------------------
enabling rules from sg2-ky:
[...]

PR #261, opened 2023-12-10
42 explanability 1

ShiriMoran:
first step in solution for 
https://github.com/np-guard/vpc-network-config-analyzer/issues/42
Specifically, simple required functionality added in both ends of the solution.

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -106,21 +106,21 @@ func _main(cmdlineArgs []string) error {
 		return fmt.Errorf(ErrorFormat, ParsingErr, err)
 	}
 
-	rc, err := ibmvpc.ParseResourcesFromFile(*inArgs.InputConfigFile)
-	if err != nil {
+	rc, err1 := ibmvpc.ParseResourcesFromFile(*inArgs.InputConfigFile)
+	if err1 != nil {
 		return fmt.Errorf("error parsing input vpc resources file: %w", err)
 	}
 
-	vpcConfigs, err := ibmvpc.VPCConfigsFromResources(rc, *inArgs.VPC, *inArgs.Debug)
-	if err != nil {
+	vpcConfigs, err2 := ibmvpc.VPCConfigsFromResources(rc, *inArgs.VPC, *inArgs.Debug)
+	if err2 != nil {
 		return fmt.Errorf(ErrorFormat, InGenerationErr, err)
 	}
 
 	// todo: tmp
 	if tmpForExplainabilityMode {
 		for _, vpcConfig := range vpcConfigs {
-			err := vpcConfig.ExplainConnectivity("vsi2-ky[10.240.20.4]", "vsi1-ky[10.240.10.4]")
-			if err != nil {
+			err3 := vpcConfig.ExplainConnectivity("vsi2-ky[10.240.20.4]", "vsi1-ky[10.240.10.4]")
+			if err3 != nil {
 				fmt.Printf("error: %s\n", err.Error())
 			}
 			return nil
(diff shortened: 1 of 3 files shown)


cmd/analyzer/main.go
```diff
@@ -104,16 +106,27 @@ func _main(cmdlineArgs []string) error {
...
+	if tmpForExplainabilityMode {
+		for _, vpcConfig := range vpcConfigs {
+			err3 := vpcConfig.ExplainConnectivity("vsi2-ky[10.240.20.4]", "vsi1-ky[10.240.10.4]")
+			if err3 != nil {
+				fmt.Printf("error: %s\n", err.Error())
+			}
+			return nil
+		}
+	}
+
```
adisos, line 129: Instead of having tmp code here, can try to have unit tests instead?

pkg/vpcmodel/nodesExplainability..go
```diff
@@ -0,0 +1,64 @@
...
+//    In this case after ParseResourcesFromFile and VPCConfigsFromResources dedicated analysis based on ConnOfSGRules will be performed,
+//    and the general connectivity analysis functionality will not be called
+//
+//    b) Enhancing the structure
+//        type ConnectivityResult struct {
+//       by adding to it the list of the rules contributing to the connection []*SGRule/[]uniqueIdsOfSGRules
+//    In this case in addition to ParseResourcesFromFile and VPCConfigsFromResources, GetVPCNetworkConnectivity will be called
+//     after which a dedicated function for collecting the relevant SG rules will be deployed.
+//
+//   Alternative a) chosen
```
adisos, line 29: `Alternative a`  will not duplicate the code of the analysis?

cmd/analyzer/main.go
ShiriMoran: I thought the code was needed for lint, but its not. So make a long story short - code removed, once there's something non trivial to check I'll use unit tests as usual

pkg/vpcmodel/nodesExplainability..go
ShiriMoran: If I'll write the code properly then it will not.
I'll extract common parts into functions.

cmd/analyzer/main.go
```diff
@@ -104,13 +104,13 @@ func _main(cmdlineArgs []string) error {
 		return fmt.Errorf(ErrorFormat, ParsingErr, err)
 	}
 
-	rc, err := ibmvpc.ParseResourcesFromFile(*inArgs.InputConfigFile)
-	if err != nil {
+	rc, err1 := ibmvpc.ParseResourcesFromFile(*inArgs.InputConfigFile)
+	if err1 != nil {
 		return fmt.Errorf("error parsing input vpc resources file: %w", err)
```
adisos, line 109:
```suggestion
		return fmt.Errorf("error parsing input vpc resources file: %w", err1)
```

cmd/analyzer/main.go
```diff
@@ -104,13 +104,13 @@ func _main(cmdlineArgs []string) error {
...
 	}
 
-	vpcConfigs, err := ibmvpc.VPCConfigsFromResources(rc, *inArgs.VPC, *inArgs.Debug)
-	if err != nil {
+	vpcConfigs, err2 := ibmvpc.VPCConfigsFromResources(rc, *inArgs.VPC, *inArgs.Debug)
+	if err2 != nil {
 		return fmt.Errorf(ErrorFormat, InGenerationErr, err)
```
adisos, line 114:
```suggestion
		return fmt.Errorf(ErrorFormat, InGenerationErr, err2)
```

adisos (approved):

PR #283, opened 2023-12-19
Issue 134 - consume from the collector the ResourcesContainter

adisos:


pkg/ibmvpc/parser.go
```diff
@@ -1027,12 +789,18 @@ func getSubnetByCidr(m map[string]*Subnet, cidr string) (*Subnet, error) {
...
-		subnet, err := getSubnetByCidr(subnetNameToSubnet, iksNode.Cidr)
+	rc *datamodel.ResourcesContainerModel,
+	skipByVPC func(string) bool) error {
+	for _, iksNode := range rc.IKSWorkerNodes {
+		if len(iksNode.NetworkInterfaces) != 1 {
+			return errIksParsing
+		}
+		iksNodeNetIntfObj := iksNode.NetworkInterfaces[0]
+
+		subnet, err := getSubnetByCidr(subnetNameToSubnet, *iksNodeNetIntfObj.Cidr)
```
zivnevo, line 800: Question: should we be concerned that subnet name is not unique across VPCs? Should we better use CRN?

zivnevo (approved):
This is so much better!!!
Thanks for this great work.
Please see my question below (if this is a problem, open an issue and handle in another PR)

pkg/ibmvpc/parser.go
adisos: right, opened issue #324

PR #286, opened 2023-12-20
281 explainability default

ShiriMoran:


pkg/ibmvpc/sg_analysis.go
```diff
@@ -11,10 +11,13 @@ import (
 	vpcmodel "github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
 )
 
+const allCidr = "0.0.0.0/0"
+
 type SGAnalyzer struct {
 	sgResource          *vpc1.SecurityGroup
 	ingressRules        []*SGRule
 	egressRules         []*SGRule
+	rulesAreDefault     bool // rules are the default ones
```
haim-kermany, line 20: the name rulesAreDefault suggest that there are more than one rule, and they are all default.
1. consider renaming
2. also rename areRulesDefault()
3. IsDefault() has basically the same meaning, consider give all the same name

pkg/ibmvpc/vpc.go
```diff
@@ -395,12 +399,25 @@ func (sgl *SecurityGroupLayer) RulesInConnectivity(src, dst vpcmodel.Node,
...
 
+// IsDefault the layer is considered default if all sgs' rules are default
+func (sgl *SecurityGroupLayer) IsDefault() bool {
+	isDefault := true
+	for _, sg := range sgl.sgList {
+		isDefault = isDefault && sg.analyzer.rulesAreDefault
+	}
+	return isDefault
+}
+
```
haim-kermany, line 410:
```suggestion
func (sgl *SecurityGroupLayer) IsDefault() bool {
	for _, sg := range sgl.sgList {
		if !sg.analyzer.rulesAreDefault{
		    return false
		}
	}
	return true
}
```
faster? cleaner?

pkg/ibmvpc/vpc.go
```diff
@@ -395,12 +399,25 @@ func (sgl *SecurityGroupLayer) RulesInConnectivity(src, dst vpcmodel.Node,
 	return res, nil
 }
 
+// IsDefault the layer is considered default if all sgs' rules are default
+func (sgl *SecurityGroupLayer) IsDefault() bool {
```
haim-kermany, line 403: when referring to a method in documentation, please add () (should be IsDefault() )
fix also in other other methods documentation

pkg/ibmvpc/sg_analysis.go
```diff
@@ -11,10 +11,13 @@ import (
 	vpcmodel "github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
 )
 
+const allCidr = "0.0.0.0/0"
```
haim-kermany, line 14: use the definition from common/ipBlock.go

haim-kermany (changes requested):
some minor comments

pkg/ibmvpc/sg_analysis.go
ShiriMoran: The issue here is that there is a default `SG`, and this is not what we are checking here. 
Also,  giving the same name to a field and a method drives it crazy.
Names changed, to avoid confusion I added comments.

pkg/ibmvpc/vpc.go
ShiriMoran: my GoLand editor insists I add the function name without the ()
I agree with you, but never messes with GoLand

pkg/ibmvpc/sg_analysis.go
```diff
@@ -288,9 +290,28 @@ func (sga *SGAnalyzer) prepareAnalyzer(sgMap map[string]*SecurityGroup, currentS
...
+		return false
+	}
+	egressRule := sga.egressRules[0]
+	egressRuleCidrs := egressRule.target.ToCidrList()
+	if len(egressRuleCidrs) != 1 {
+		return false
+	}
+	if egressRuleCidrs[0] == common.CidrAll {
+		return true
+	}
```
adisos, line 311: should also check that the rule's connection is "all"

pkg/vpcmodel/abstractVPC.go
```diff
@@ -89,6 +89,8 @@ type FilterTrafficResource interface {
 	// RulesInConnectivity get the list of rules of a given filter that contributes to the connection between src and dst
 	// todo: currently implemented only to sg; likely src and dst will be VPCResourceIntf instead of Node
 	RulesInConnectivity(src, dst Node, isIngress bool) ([]RulesInFilter, error)
+	// IsDefault returns true iff all sgs/nacls/.. are the default ones
+	IsDefault() bool
```
adisos, line 93: Why do we need to expose `IsDefault()` in the level of SGLayer or NACLLayer? 
In the explainability context it is relevant to check for a certain resource(s) involved in the two input nodes connectivity only, and not for the entire layer of all such resources..

pkg/ibmvpc/sg_analysis.go
ShiriMoran: right

pkg/vpcmodel/abstractVPC.go
ShiriMoran: Not sure. We might want to inform the user in the future that all relevant rules are default. Anyways, since the above functions are not actually used at the moment I removed them.

adisos (approved):

haim-kermany (approved):

PR #293, opened 2023-12-24
remove duplicated message about unidriectional connections

olasaadi99:


pkg/ibmvpc/analysis_output_test.go
adisos, line 1: why is this change required on the test functions?

pkg/ibmvpc/examples/demo_with_instances.json
adisos, line 1: why is the JSON output impacted by this PR?

adisos (commented):
Looks good, added a few comments.

pkg/ibmvpc/examples/demo_with_instances.json
olasaadi99: because I am now comparing the after aggregate always
before the change, the comparision was after aggregate only if number of headers > 1, and there is a diff between the file with one header after aggregate and before

pkg/ibmvpc/examples/acl_testing3.json
adisos, line 1: why is this file removed from the tests output files? same for `sg_testing1.json` and for `sg_testing1_new.json`?

olasaadi99: it was deleted because there is no test for it.. (I deleted all of the generated outputs and generated new ones with the new names)

adisos (approved):

Issue #237, opened 2023-11-27
inconsistent representation of network interfaces


ShiriMoran:
The representation of network interfaces is not consistent.
Specifically, while its name is indeed the name of the network interface, the `id` refereed to as `ResourceUID`  is taken from the network interface's `primary_ip`.
This is visible in the `json` format.

E.g.:
npGurard output include:
```
                "dst": {
                    "ResourceName": "yarn-canary-guileless-deftly",
                    "ResourceUID": "id:21",
                    "ResourceType": "NetworkInterface",
                    "Zone": "us-south-1"
                },
```
but the source is the following:
```
                    "href": "href:18",
                    "id": "id:19",
                    "name": "yarn-canary-guileless-deftly",
                    "port_speed": 3000,
                    "primary_ip": {
                        "address": "10.240.20.4",
                        "href": "href:20",
                        "id": "id:21",
                        "name": "precision-grudge-daylight-married",
                        "resource_type": "subnet_reserved_ip"
                    },
```

adisos:
The issue is that `SG` targets use the network interface's name and id , whereas the `FIP` uses the `id` of the network interface's `primary_ip`

PR #298, opened 2023-12-27
fix issue #237 : use network interface ID instead of PrimaryIP.ID

adisos:


pkg/ibmvpc/parser.go
```diff
@@ -447,7 +447,7 @@ func getInstancesConfig(
 			res[vpcUID].Nodes = append(res[vpcUID].Nodes, intfNode)
 			res[vpcUID].NameToResource[intfNode.Name()] = intfNode
 			vsiNode.nodes = append(vsiNode.nodes, intfNode)
-			intfNameToIntf[*netintf.Name] = intfNode
+			intfNameToIntf[intfNode.ResourceName] = intfNode
```
ShiriMoran, line 450: This change seems to have no effect; the two properties are identical. The point is just referring to the original property? 
Its not specifically related to #237, right?

ShiriMoran (approved):

PR #301, opened 2023-12-31
290 explainability public ip grouping

ShiriMoran:


pkg/vpcmodel/grouping.go
```diff
@@ -85,13 +88,26 @@ func newGroupConnLinesDiff(d *diffBetweenCfgs) (res *GroupConnLines, err error)
...
 
+func newGroupConnExplainability(c *VPCConfig, e *explainStruct) (res *GroupConnLines, err error) {
+	res = &GroupConnLines{
+		c:                        c,
+		e:                        e,
+		srcToDst:                 newGroupingConnections(),
+		dstToSrc:                 newGroupingConnections(),
+		groupedEndpointsElemsMap: make(map[string]*groupedEndpointsElems),
+		groupedExternalNodesMap:  make(map[string]*groupedExternalNodes)}
+	err = res.groupExternalAddressesForExplainability()
```
haim-kermany, line 99:
```suggestion
	groupedEndpointsElemsMap: map[string]*groupedEndpointsElems{},
		groupedExternalNodesMap:  map[string]*groupedExternalNodes{}
	err = res.groupExternalAddressesForExplainability()
```

haim-kermany (dismissed):

pkg/vpcmodel/grouping.go
ShiriMoran: done

pkg/vpcmodel/grouping.go
```diff
@@ -282,23 +301,40 @@ func (g *GroupConnLines) groupExternalAddressesForDiff(thisMinusOther bool) erro
...
-func (g *GroupConnLines) addLineToExternalGrouping(res *[]*groupedConnLine, emptyConn bool,
-	src, dst VPCResourceIntf, commonProps *groupedCommonProperties) error {
-	if emptyConn {
-		return nil
+// group public internet ranges for explainability lines
+func (g *GroupConnLines) groupExternalAddressesForExplainability() error {
+	var res []*groupedConnLine
+	for _, rulesSrcDst := range *g.e {
+		connStr := ""
+		if rulesSrcDst.conn != nil {
```
adisos, line 322: can `rulesSrcDst.conn` be `nil`? is it valid?

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -24,6 +26,13 @@ type rulesSingleSrcDst struct {
 
 type explainStruct []*rulesSingleSrcDst
 
+type explanation struct {
+	c             *VPCConfig
+	explainStruct *explainStruct
+	// grouped connectivity result
```
adisos, line 32:
```suggestion
	// grouped connectivity result -- grouping common explanation lines with common src/dst (internal node) and different dst/src (external node)  [required due to computation with disjoint ip-blocks] 
```

pkg/vpcmodel/grouping.go
```diff
@@ -85,13 +88,26 @@ func newGroupConnLinesDiff(d *diffBetweenCfgs) (res *GroupConnLines, err error)
 	return res, err
 }
 
+func newGroupConnExplainability(c *VPCConfig, e *explainStruct) (res *GroupConnLines, err error) {
+	res = &GroupConnLines{
+		c:                        c,
+		e:                        e,
```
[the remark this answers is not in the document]
ShiriMoran: done, also for s, d and v

pkg/vpcmodel/grouping.go
ShiriMoran: Yes. It can be valid in cases there is no connection. These cases are still relevant for the explanation and should be grouped.

adisos: why using `nil` instead of empty ConnectionSet object?

ShiriMoran: Done.

pkg/vpcmodel/grouping.go
```diff
@@ -275,30 +294,47 @@ func (g *GroupConnLines) groupExternalAddressesForDiff(thisMinusOther bool) erro
...
-	if emptyConn {
-		return nil
+// group public internet ranges for explainability lines
+func (g *GroupConnLines) groupExternalAddressesForExplainability() error {
+	var res []*groupedConnLine
+	for _, rulesSrcDst := range *g.explain {
+		connStr := ""
+		if !rulesSrcDst.conn.IsEmpty() {
+			connStr = rulesSrcDst.conn.String() + semicolon
+		}
```
adisos, line 324:
```suggestion
		connStr := rulesSrcDst.conn.String() + semicolon
```
can also use `conn.String()` for an empty connection-set (string is "no connections")

adisos (approved):

PR #310, opened 2024-01-04
302 explainabilty query connection

ShiriMoran:


ShiriMoran:
I need to fix something; closing the PR until I do, will reopen it once ready

PR #311, opened 2024-01-04
Better zone indexes

haim-kermany:
improving algorithm for setting the zone indexes.
for cases in which there the zones can be spits to subsets that has no grouping connection
(common case - zones from different VPCs)

### DIFF
--- a/pkg/drawio/subnetsLayout.go
+++ b/pkg/drawio/subnetsLayout.go
@@ -470,7 +470,7 @@ func (ly *subnetsLayout) calcZoneOrder() {
 	for len(zonesScores) > 0 {
 		zoneToAdd, addToRight, newZoneOrder := chooseZoneToAdd(zonesScores, zoneOrder)
 		if newZoneOrder {
-			if zoneOrder != nil {
+			if len(zoneOrder) > 0 {
 				zoneOrders = append(zoneOrders, zoneOrder)
 			}
 			zoneOrder = []TreeNodeInterface{}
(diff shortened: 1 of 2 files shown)


pkg/drawio/subnetsLayout.go
```diff
@@ -465,9 +465,16 @@ func (ly *subnetsLayout) newGroupFromSplitMiniGroups(group *groupDataS, miniGrou
...
 func (ly *subnetsLayout) calcZoneOrder() {
 	zonesScores := ly.calcZonePairScores()
-	zoneOrder := []TreeNodeInterface{}
+	var zoneOrder []TreeNodeInterface
+	zoneOrders := [][]TreeNodeInterface{}
 	for len(zonesScores) > 0 {
-		zoneToAdd, addToRight := chooseZoneToAdd(zonesScores, zoneOrder)
+		zoneToAdd, addToRight, newZoneOrder := chooseZoneToAdd(zonesScores, zoneOrder)
+		if newZoneOrder {
+			if zoneOrder != nil {
```
zivnevo, line 473:
```suggestion
			if len(zoneOrder) > 0 {
```
Looks safer, and consistent with the condition after the end of the loop

zivnevo (approved):
I have just this tiny comment. Feel free to ignore

pkg/drawio/subnetsLayout.go
haim-kermany: done

zivnevo (approved):

PR #312, opened 2024-01-07
302 explainabilty query connection

ShiriMoran:


haim-kermany (changes requested):

pkg/ibmvpc/sg_analysis.go
```diff
@@ -323,14 +323,47 @@ func (sga *SGAnalyzer) AllowedConnectivity(target string, isIngress bool) *commo
...
 	}
-	return nil
+	return nil, nil
+}
+
+// given a list of rules and a connection, return the sublist of rules that contributes to the connection
+func (sga *SGAnalyzer) getRulesRelevantConn(rules []int, conn *common.ConnectionSet) ([]int, error) {
+	relevantRules := []int{}
+	allRules := sga.ingressRules
+	allRules = append(allRules, sga.egressRules...)
```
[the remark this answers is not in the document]
ShiriMoran: Done. Thanks.

pkg/ibmvpc/sg_analysis.go
```diff
@@ -323,14 +323,47 @@ func (sga *SGAnalyzer) AllowedConnectivity(target string, isIngress bool) *commo
...
+	for _, rule := range allRules {
+		if contains(rules, rule.index) && !conn.Intersection(rule.connections).IsEmpty() {
+			relevantRules = append(relevantRules, rule.index)
+		}
+	}
+	return relevantRules, nil
+}
+
+func contains(s []int, e int) bool {
+	for _, a := range s {
```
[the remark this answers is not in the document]
ShiriMoran: done

haim-kermany (approved):

pkg/common/genericSet.go
```diff
@@ -51,3 +51,12 @@ func (s GenericSet[T]) IsIntersect(s2 GenericSet[T]) bool {
...
 }
+
+func Contains[T comparable](s []T, e T) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
+}
```
adisos, line 62: go 1.21 has this functionality, see: https://pkg.go.dev/slices#Contains 
better use `slices.Contains()` instead of implementing it here

pkg/ibmvpc/sg_analysis.go
```diff
@@ -323,14 +323,35 @@ func (sga *SGAnalyzer) AllowedConnectivity(target string, isIngress bool) *commo
 }
 
 // rulesInConnectivity list of SG rules contributing to the connectivity
-func (sga *SGAnalyzer) rulesInConnectivity(target string, isIngress bool) []int {
+func (sga *SGAnalyzer) rulesInConnectivity(target string, conn *common.ConnectionSet, isIngress bool) ([]int, error) {
 	analyzedConns, ipb := sga.getAnalyzedConnsIPB(target, isIngress)
 	for definedTarget, rules := range analyzedConns.contribRules {
 		if ipb.ContainedIn(definedTarget) {
-			return rules
+			if conn != nil { // connection not part of the query
```
adisos, line 330:
```suggestion
			if conn != nil { // connection is part of the query
```

pkg/ibmvpc/sg_analysis.go
```diff
@@ -323,14 +323,35 @@ func (sga *SGAnalyzer) AllowedConnectivity(target string, isIngress bool) *commo
...
+				contained, err := conn.ContainedIn(analyzedConns.allowedconns[definedTarget])
```
adisos, line 331: containment of conns here and intersection of conns in `getRulesRelevantConn` -- add some documentation about the difference (here it is the top level result per target, and only if the input connection is fully contained in the allowed conn, we consider this as "allowed" and look for contributing rules, even if they only intersect and not fully contained).

pkg/ibmvpc/sg_analysis.go
```diff
@@ -323,14 +323,35 @@ func (sga *SGAnalyzer) AllowedConnectivity(target string, isIngress bool) *commo
...
+				if err != nil {
+					return nil, err
+				}
+				if contained {
+					return sga.getRulesRelevantConn(rules, conn)
+				}
+				return nil, nil
+			}
+			return rules, nil // connection not part of query
```
adisos, line 340:
```suggestion
			return rules, nil // connection not part of query - all rules are relevant
```

adisos (changes requested):

pkg/ibmvpc/sg_analysis.go
ShiriMoran: done

pkg/common/genericSet.go
ShiriMoran: indeed! done.

adisos (approved):

PR #319, opened 2024-01-11
accept a description of a connection in the cli and translate it to co…

olasaadi99:
…mmon.ConnectionSet

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -86,10 +86,11 @@ func vpcConfigsFromFile(fileName string, inArgs *InArgs) (map[string]*vpcmodel.V
 
 func translateCDtoConnectionSet(inArgs *InArgs) *common.ConnectionSet {
 	connection := common.NewConnectionSet(false)
-	if inArgs.Protocol == "icmp" {
-		connection.AddICMPConnection(inArgs.SrcMinPort, inArgs.SrcMaxPort, common.MinPort, common.MaxPort)
+	if *inArgs.QProtocol == "ICMP" {
+		connection.AddICMPConnection(*inArgs.QSrcMinPort, *inArgs.QSrcMaxPort, *inArgs.QDstMinPort, *inArgs.QDstMaxPort)
+	} else {
+		connection.AddTCPorUDPConn(common.ProtocolStr(*inArgs.QProtocol), *inArgs.QSrcMinPort, *inArgs.QSrcMaxPort, *inArgs.QDstMinPort, *inArgs.QDstMaxPort)
 	}
-	connection.AddTCPorUDPConn(common.ProtocolStr(inArgs.Protocol), inArgs.SrcMinPort, inArgs.SrcMaxPort, common.MinPort, common.MaxPort)
 
 	return connection
 }
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -137,7 +137,10 @@ func _main(cmdlineArgs []string) error {
 	}
 	fmt.Println(vpcAnalysisOutput)
 
-	translateCDtoConnectionSet(inArgs)
+	if *inArgs.QProtocol != "" {
+		_ = translateCDtoConnectionSet(inArgs)
+	}
+
 	return nil
 }
 
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -49,6 +49,8 @@ func analysisTypeToUseCase(inArgs *InArgs) vpcmodel.OutputUseCase {
 		return vpcmodel.SubnetsDiff
 	case allEndpointsDiff:
 		return vpcmodel.EndpointsDiff
+	case queryMode:
+		return vpcmodel.Query
 	}
 	return vpcmodel.AllEndpoints
 }
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -86,10 +86,11 @@ func vpcConfigsFromFile(fileName string, inArgs *InArgs) (map[string]*vpcmodel.V
 
 func translateCDtoConnectionSet(inArgs *InArgs) *common.ConnectionSet {
 	connection := common.NewConnectionSet(false)
-	if *inArgs.QProtocol == "ICMP" {
+	if common.ProtocolStr(*inArgs.QProtocol) == common.ProtocolICMP {
 		connection.AddICMPConnection(*inArgs.QSrcMinPort, *inArgs.QSrcMaxPort, *inArgs.QDstMinPort, *inArgs.QDstMaxPort)
 	} else {
-		connection.AddTCPorUDPConn(common.ProtocolStr(*inArgs.QProtocol), *inArgs.QSrcMinPort, *inArgs.QSrcMaxPort, *inArgs.QDstMinPort, *inArgs.QDstMaxPort)
+		connection.AddTCPorUDPConn(common.ProtocolStr(*inArgs.QProtocol), *inArgs.QSrcMinPort, *inArgs.QSrcMaxPort,
+			*inArgs.QDstMinPort, *inArgs.QDstMaxPort)
 	}
 
 	return connection
@@ -135,6 +136,8 @@ func _main(cmdlineArgs []string) error {
 		return err2
 	}
 	fmt.Println(vpcAnalysisOutput)
+
+	translateCDtoConnectionSet(inArgs)
 	return nil
 }
 
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -49,8 +49,8 @@ func analysisTypeToUseCase(inArgs *InArgs) vpcmodel.OutputUseCase {
 		return vpcmodel.SubnetsDiff
 	case allEndpointsDiff:
 		return vpcmodel.EndpointsDiff
-	case queryMode:
-		return vpcmodel.Query
+	case explainMode:
+		return vpcmodel.Explain
 	}
 	return vpcmodel.AllEndpoints
 }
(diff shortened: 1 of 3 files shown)


cmd/analyzer/main.go
```diff
@@ -83,6 +84,16 @@ func vpcConfigsFromFile(fileName string, inArgs *InArgs) (map[string]*vpcmodel.V
 	return vpcConfigs, nil
 }
 
+func translateCDtoConnectionSet(inArgs *InArgs) *common.ConnectionSet {
+	connection := common.NewConnectionSet(false)
+	if inArgs.Protocol == "icmp" {
+		connection.AddICMPConnection(inArgs.SrcMinPort, inArgs.SrcMaxPort, common.MinPort, common.MaxPort)
```
ShiriMoran, line 90: Please have the `portMin` and `portMax` for `src` and `dst` be potential part of the input; if it is left unspecified than use the `min` and `max` values as you do now.

cmd/analyzer/parse_args.go
```diff
@@ -18,6 +21,10 @@ type InArgs struct {
 	VPC                   *string
 	Debug                 *bool
 	Version               *bool
+	ConnectionDescription *string
+	Protocol              string
+	SrcMinPort            int64
+	SrcMaxPort            int64
 }
```
ShiriMoran, line 28: Please add min and max port also for dst
Also, consider adding Q before these new properties or something else so that it will be clear these are relevant only in query mode

cmd/analyzer/parse_args.go
```diff
@@ -18,6 +21,10 @@ type InArgs struct {
 	VPC                   *string
 	Debug                 *bool
 	Version               *bool
+	ConnectionDescription *string
```
ShiriMoran, line 24: Why is the `ConnectionDescription` required?

cmd/analyzer/parse_args.go
```diff
@@ -160,6 +169,8 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
 	args.VPC = flagset.String(VPC, "", "CRN of the VPC to analyze")
 	args.Debug = flagset.Bool(Debug, false, "Run in debug mode")
 	args.Version = flagset.Bool(Version, false, "Prints the release version number")
+	args.ConnectionDescription = flagset.String(ConnectionDescription, "",
+		"Description to a connection in the format: 'Protocol', 'MinSrc', 'MaxSrc'")
```
ShiriMoran, line 173: Again, why is this required?

cmd/analyzer/parse_args.go
```diff
@@ -180,9 +191,48 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
+	err = parseConnectionDescription(args.ConnectionDescription, &args)
+	if err != nil {
+		return nil, err
+	}
 
 	return &args, nil
 }
+
+func parseConnectionDescription(connectionDescription *string, args *InArgs) error {
+	if *connectionDescription == "" {
```
ShiriMoran, line 203: This is not needed. We will soon add `AnalysisType` for query, and then allow specifying these parms only when the mode will be query.
In the meantime just have validity checks that if protocol is specified is legal and that if src/dst min/max ports are specified they are withing range, and that if both min and max are specified then the min is lower than or equal to max

ShiriMoran (changes requested):

cmd/analyzer/main.go
```diff
@@ -83,6 +84,16 @@ func vpcConfigsFromFile(fileName string, inArgs *InArgs) (map[string]*vpcmodel.V
 	return vpcConfigs, nil
 }
 
+func translateCDtoConnectionSet(inArgs *InArgs) *common.ConnectionSet {
+	connection := common.NewConnectionSet(false)
+	if inArgs.Protocol == "icmp" {
```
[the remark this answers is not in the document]
olasaadi99: it is supported, I convert it to lower case in parse args..

cmd/analyzer/parse_args.go
olasaadi99: though it will be easier for the user to pass single input for the connection description instead of each param (and I can parse it)

ShiriMoran: I see. Perhaps I mislead you. Its better, I think, to use the existing mechanism for feeding parms. This why we can use the existing parsing and validity mechanism. E.g,use the go args mechanism to enforce that ports parm are ints withing range

cmd/analyzer/main.go
```diff
@@ -83,6 +84,18 @@ func vpcConfigsFromFile(fileName string, inArgs *InArgs) (map[string]*vpcmodel.V
 	return vpcConfigs, nil
 }
 
+func translateCDtoConnectionSet(inArgs *InArgs) *common.ConnectionSet {
+	connection := common.NewConnectionSet(false)
+	if common.ProtocolStr(*inArgs.QProtocol) == common.ProtocolICMP {
```
ShiriMoran, line 89: Please convert to upper case when comparing with  e.g. common.ProtocolICMP

cmd/analyzer/main.go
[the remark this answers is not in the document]
ShiriMoran: But now in the fixed version you convert it to upper case?

ShiriMoran:
Much better :-)
To make the parms complete, please add to `AnalysisType` another possible value `query` and add the following validity checks:
1. protocol must be specified for  `AnalysisType` query
2. All the new parms can be specified only if `AnalysisType` is query

cmd/analyzer/main.go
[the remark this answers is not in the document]
olasaadi99: right

cmd/analyzer/parse_args.go
```diff
@@ -180,9 +205,74 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
 	}
+	err = invalidArgsQueryMode(&args, flagset)
+	if err != nil {
+		return nil, err
+	}
 
 	return &args, nil
 }
+
+func isFlagPassed(name string, flagset *flag.FlagSet) bool {
```
ShiriMoran, line 216: Perhaps remain to `wasFlagSpecified`

cmd/analyzer/parse_args.go
```diff
@@ -180,9 +205,74 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
+	found := false
+	flagset.Visit(func(f *flag.Flag) {
+		if f.Name == name {
+			found = true
+		}
+	})
+	return found
+}
+
+func isQueryModeParamsPassed(flagset *flag.FlagSet) bool {
```
ShiriMoran, line 226: Also here, consider renaming to `wereQueryParmSpecified`

cmd/analyzer/parse_args.go
```diff
@@ -180,9 +205,74 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
+
+	if *args.QDstMinPort > *args.QDstMaxPort {
+		return fmt.Errorf("DstMaxPort %d should be higher than DstMinPort %d", *args.QSrcMaxPort, *args.QSrcMinPort)
+	}
+
+	if *args.QDstMinPort > common.MaxPort || *args.QDstMinPort < common.MinPort ||
+		*args.QDstMaxPort > common.MaxPort || *args.QDstMaxPort < common.MinPort {
+		return fmt.Errorf("DstMaxPort and DstMinPort must be in ranges [%d, %d]", common.MinPort, common.MaxPort)
+	}
+
```
ShiriMoran, line 253: Please consider having the checks for `dst/src` in a function called once, to avoid code duplication

ShiriMoran:
Looks good, a few more minor comments

ShiriMoran (changes requested):

pkg/vpcmodel/output.go
```diff
@@ -38,6 +38,7 @@ const (
 	AllSubnetsNoPGW                      // connectivity between subnets (consider nacl only)
 	SubnetsDiff                          // diff between subnets connectivity of two cfgs (consider nacl + pgw)
 	EndpointsDiff                        // diff between vsis connectivity of two cfgs
+	Query                                // analyze a connection description
```
adisos, line 41:
```suggestion
	Explain                                // explain specified connectivity, given src,dst and connection 
```

cmd/analyzer/parse_args.go
```diff
@@ -59,6 +76,7 @@ const (
 	singleSubnet     = "single_subnet"      // single subnet connectivity analysis
 	allEndpointsDiff = "diff_all_endpoints" // semantic diff of allEndpoints analysis between two configurations
 	allSubnetsDiff   = "diff_all_subnets"   // semantic diff of allSubnets analysis between two configurations
+	queryMode        = "query"              // analyze a connection description
```
adisos, line 79:
```suggestion
	explainMode        = "explain"              // explain specified connectivity
```

adisos (changes requested):
Please rename "query" analysis type to be "explain" instead (@ShiriMoran  confirmed the renaming does make sense).
Added few examples in comments where to apply this change, go over all relevant places to change.

cmd/analyzer/main.go
```diff
@@ -83,6 +86,18 @@ func vpcConfigsFromFile(fileName string, inArgs *InArgs) (map[string]*vpcmodel.V
...
+		connection.AddICMPConnection(*inArgs.QSrcMinPort, *inArgs.QSrcMaxPort, *inArgs.QDstMinPort, *inArgs.QDstMaxPort)
```
adisos, line 92: why are src/dst ports sent to `AddICMPConnection` as icmp type and code? 
initially for simplicity, we do not have input icmp type / code, so should add ALL ICMP connections (use the defualt min and max type/code values from the `common` pkg).

cmd/analyzer/parse_args.go
```diff
@@ -180,9 +205,78 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
+func PortInRange(port int64) bool {
+	if port > common.MaxPort || port < common.MinPort {
+		return false
+	}
+
+	return true
+}
+
+func validRangeConnectionExplainMode(args *InArgs) error {
+	if *args.QSrcMinPort > *args.QSrcMaxPort {
```
ShiriMoran, line 244: minPort and maxPort may be equal

cmd/analyzer/parse_args.go
```diff
@@ -180,9 +205,78 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
+		return fmt.Errorf("srcMaxPort %d should be higher than srcMinPort %d", *args.QSrcMaxPort, *args.QSrcMinPort)
+	}
+
+	if *args.QDstMinPort > *args.QDstMaxPort {
```
ShiriMoran, line 248: Also here, minPort and maxPort may be equal

ShiriMoran (changes requested):
Almost done.
Just one comment: please let me mark as resolve my comments, so that I can easily follow them

ShiriMoran (approved):

PR #323, opened 2024-01-15
add tgw example with collection from all regions

adisos:


kyorav (approved):
The results look good to me, did not review code.

PR #355, opened 2024-01-31
improve unit tests

olasaadi99:


pkg/ibmvpc/examples/out/analysis_out/actual.txt
```diff
@@ -0,0 +1,15 @@
+Connectivity explanation between vsi1-ky[10.240.10.4] and 161.26.0.0/16
```
ShiriMoran, line 1: This file should not be part of the PR

pkg/ibmvpc/examples/out/analysis_out/expected.txt
```diff
@@ -0,0 +1,20 @@
+Connectivity explanation for protocol: TCP between vsi1-ky[10.240.10.4] and 161.26.0.0/16
```
ShiriMoran, line 1: This file should not be part of the PR

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -139,7 +139,9 @@ func getTestFileSuffix(format vpcmodel.OutFormat) (suffix string, err error) {
 // initTest: based on the test name, set the input config file name, and the output
 // files names (actual and expected), per use case
 func (tt *vpcGeneralTest) initTest() {
-	tt.inputConfig = inputFilePrefix + tt.name + ".json"
+	if tt.inputConfig == "" {
```
ShiriMoran, line 142: what is the scenario in which `tt.inputConfig` is empty?

pkg/ibmvpc/explainability_test.go
```diff
@@ -28,544 +28,300 @@ func getConfig(t *testing.T, fileName string) *vpcmodel.VPCConfig {
...
-	EDstMaxPort int64
-	out         string
-}
-
-var explainTests = []*explainGeneralTest{
+var explainTests = []*vpcGeneralTest{
 	{
 		name:        "VsiToVsi1",
-		inputConfig: "input_sg_testing1_new",
+		inputConfig: "input_sg_testing1_new.json",
```
ShiriMoran, line 34: Why is the suffix `*json` required

ShiriMoran (commented):
Much much better.
Few comments in the files
Please add a function of generating the current explainability results, similarly to `TestAllWithGeneration`  (and based on the same code)

pkg/ibmvpc/analysis_output_test.go
olasaadi99: in analysis_output_tests we do not pass the input_config.. we get it from name(which is uniqe for each test) but in explain tests it is not

pkg/ibmvpc/explainability_test.go
olasaadi99: because we get the input file from here

ShiriMoran: What I meant is why can't we do without the suffix and add it automatically as in `analysis_output_test.go`

ShiriMoran (commented):
@olasaadi99 I believe the files `ibmvpc/out_*` can be deleted

ShiriMoran (dismissed):

pkg/ibmvpc/analysis_output_test.go
adisos: can this be changed to be unique for both `analysis_output_tests` and `explainability` tests? (not necessarily in this PR, can do this as a followup).

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -570,7 +572,7 @@ func getVPCConfigs(t *testing.T, tt *vpcGeneralTest, firstCfg bool) map[string]*
 	} else {
 		inputConfig = tt.inputConfig2nd
 	}
-	inputConfigFile := filepath.Join(getTestsDir(), inputConfig)
+	inputConfigFile := filepath.Join(getTestsDir("input"), inputConfig)
```
adisos, line 575: better have a separate function instead of repeating `filepath.Join(getTestsDir("input"), ...` with hardcoded `input` dir prefix (and also define it as const)

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -654,8 +658,8 @@ func cleanStr(str string) string {
 // compareTextualResult is called in case of output mismatch, to provide more details on the difference
 func compareTextualResult(expected, actual string) {
 	var err1, err2 error
-	_, err1 = vpcmodel.WriteToFile(expected, filepath.Join(getTestsDir(), "expected.txt"))
-	_, err2 = vpcmodel.WriteToFile(actual, filepath.Join(getTestsDir(), "actual.txt"))
+	_, err1 = vpcmodel.WriteToFile(expected, filepath.Join(getTestsDir("out/analysis_out"), "expected.txt"))
+	_, err2 = vpcmodel.WriteToFile(actual, filepath.Join(getTestsDir("out/analysis_out"), "actual.txt"))
```
adisos, line 662: same here for `filepath.Join(getTestsDir("out/analysis_out"), ..."`

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -677,7 +681,7 @@ func compareTextualResult(expected, actual string) {
 }
 
 // getTestsDir returns the path to the dir where test input and output files are located
-func getTestsDir() string {
+func getTestsDir(dir string) string {
 	currentDir, _ := os.Getwd()
-	return filepath.Join(currentDir, "examples")
+	return filepath.Join(currentDir, "examples/"+dir)
```
adisos, line 686: also `"examples"` should be a const

pkg/ibmvpc/explainability_test.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // getConfigs returns  map[string]*vpcmodel.VPCConfig obj for the input test (config json file)
 func getConfig(t *testing.T, fileName string) *vpcmodel.VPCConfig {
-	inputConfigFile := filepath.Join(getTestsDir(), fileName+".json")
+	inputConfigFile := filepath.Join(getTestsDir("input"), fileName+".json")
```
adisos, line 16: same here

pkg/ibmvpc/explainability_test.go
adisos: I also think we should try to be consistent

pkg/ibmvpc/naclAnalysis_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestGetRules(t *testing.T) {
-	rc, err := ParseResourcesFromFile(filepath.Join(getTestsDir(), "input_acl_testing3.json"))
+	rc, err := ParseResourcesFromFile(filepath.Join(getTestsDir("input"), "input_acl_testing3.json"))
```
adisos, line 14: same comment here

olasaadi99:
-> I also think we should try to be consistent
              in analysisOut test we get input config from the name  of the test (and it should not be with .json because it is a name) 
              and add ".json" later
              in explain test the config file is not uniqe for each test therfore we can not name the test based on the config file that is 
              why I added this line to check if the config file is empty string

adisos:
> -> I also think we should try to be consistent in analysisOut test we get input config from the name of the test (and it should not be with .json because it is a name) and add ".json" later in explain test the config file is not uniqe for each test therfore we can not name the test based on the config file that is why I added this line to check if the config file is empty string

you can change the attritbue `name` to be `config_name` in both tests variants, for which can have consistent handling

olasaadi99:
so in analysisOut test there will be name("testFile") nad config file ("testFile.json") right?

olasaadi99:
> > -> I also think we should try to be consistent in analysisOut test we get input config from the name of the test (and it should not be with .json because it is a name) and add ".json" later in explain test the config file is not uniqe for each test therfore we can not name the test based on the config file that is why I added this line to check if the config file is empty string
> 
> you can change the attritbue `name` to be `config_name` in both tests variants, for which can have consistent handling

so in analysisOut test there will be name("testFile") nad config file ("testFile.json") right?

adisos:
> > > -> I also think we should try to be consistent in analysisOut test we get input config from the name of the test (and it should not be with .json because it is a name) and add ".json" later in explain test the config file is not uniqe for each test therfore we can not name the test based on the config file that is why I added this line to check if the config file is empty string
> > 
> > 
> > you can change the attritbue `name` to be `config_name` in both tests variants, for which can have consistent handling
> 
> so in analysisOut test there will be name("testFile") nad config file ("testFile.json") right?

I think we can avoid duplicating the args in `analysis_out`.. why not just renaming the `name` arg to be `config_name`? and also have such attribute in `explain` tests? 
The test name can also be based on this arg for the `analysis_out` test.

adisos (commented):
why are some of the output files marked as deleted and not as moved?

adisos:
> why are some of the output files marked as deleted and not as moved?

seems only related to old files that compared output per vpc, not relevant anymore

adisos (approved):
let's merge, and have a follow-up PR for improvements of a more consistent handling

PR #361, opened 2024-02-01
overlap for src-dest with the same center in drawio

haim-kermany:
handling the case of one square with two self loop.
in general, when there is a self loop, drawio adds point, and the line looks OK
however, when there are two self loops, drawio draw them one over the other.


zivnevo:
@haim-kermany , can u please explain how such a case can happy in real life? If I understand correctly, the connectivity from a square to itself will always be represented by a single connection set.

haim-kermany:
> @haim-kermany , can u please explain how such a case can happy in real life? If I understand correctly, the connectivity from a square to itself will always be represented by a single connection set.

That is a good question.
I should probably asked it on Thursday morning

pkg/drawio/layout.go
```diff
@@ -26,6 +26,9 @@ import (
 
 const (
 	minSize         = 10
+	twoMinSize      = 20
+	threeMinSize    = 20
```
zivnevo, line 30: `twoMinSize` and `threeMinSize` are not being used.

pkg/drawio/layout.go
```diff
@@ -26,6 +26,9 @@ import (
...
+	fourMinSize     = 40
```
zivnevo, line 31: suggest to rename to `fourTimesMinSize`

zivnevo (changes requested):

pkg/drawio/layout.go
haim-kermany: fixed

pkg/drawio/layout.go
haim-kermany: fixed

zivnevo (approved):

PR #364, opened 2024-02-04
improve unit tests

olasaadi99:


ShiriMoran (approved):

ShiriMoran:
@olasaadi99 I see there are 14 dup files in explain_out:
GroupingExternalSG1_all_vpcs_explain.txt and out_GroupingExternalSG1_all_vpcs_explain.txt
and 13 others of the same type: <name> and out_<name>
Please delete the redundant files in a followup PR

olasaadi99:
these files are generated in the test (ignored with gitignore) it is to compare actual and expected (as it is in analysis out tests also)

adisos:
> these files are generated in the test (ignored with gitignore) it is to compare actual and expected (as it is in analysis out tests also)

These files should not be uploaded to the repo (maybe you should update your local gitignore so that it does not get committed)

olasaadi99:
> > these files are generated in the test (ignored with gitignore) it is to compare actual and expected (as it is in analysis out tests also)
> 
> These files should not be uploaded to the repo (maybe you should update your local gitignore so that it does not get committed)

but it is not on the repo..

ShiriMoran:
If these files are generated only for the comparison, please delete them at the end of the process

adisos:
> If these files are generated only for the comparison, please delete them at the end of the process

or there should be added `.gitignore` files to the relevant subdris you added

Issue #353, opened 2024-01-31
connect internal address to a vsi (for explainability)


ShiriMoran:
Given a string address in cidr format (e.g. `10.240.20.4/32`) or non-cidr format (e.g. `10.240.10.4`) return the list of all network interfaces attached to it (that could of course be nil)

@zivnevo do we want to support only a single IP or also a range? we can have this function support range and still at the first release allow only a single IP

zivnevo:
It's OK if we initially support only a single IP. We will likely want to expand that to a CIDR later on.

ShiriMoran:
So lets have the function handle any `CIDR` and add phase 1 restriction outside of it.
`CIDR` representing network interface of more than one VSI - I think this should be an input error, also in the future.
@zivnevo agreed?

zivnevo:
It might be very convenient for users to specify a CIDR and get (a separate? unified?) explanation for each NI with an IP in the CIDR. Users are unlikely to remember individual IPs and may not always know VSI names, but they will certainly know the subnet CIDR.
@kyorav , what do you think?

kyorav:
I think it's a good tool to have because it will let users slice the report in various ways to help answer questions like "why do users on the second floor not have access to the app when users on the first floor do".

ShiriMoran:
At first stage assume the input string presents a single internal address, e.g. `10.240.10.4` ; in 2nd stage we will handle cidrs
The signature of the function should be the following 
`func (c *VPCConfig) getNetworkInterfaces(string) ([]Node, err)`

You need to translate the string into an IP address, and then look for a Node representing an network interface that has this address.
A node representing a network interface must be internal @adisos is there a dedicated Kind() for network interfaces?

adisos:
> is there a dedicated Kind() for network interfaces?
yes, here is how it is created in `parser.go`:
```
			intfNode := &NetworkInterface{
				VPCResource: vpcmodel.VPCResource{
					ResourceName: *netintf.Name,
					ResourceUID:  *netintf.ID,
					ResourceType: ResourceTypeNetworkInterface,
					Zone:         *instance.Zone.Name,
					VPCRef:       vpc,
				},
```

ShiriMoran:
If the input string does not represent only an internal address then return error with an appropriate message

PR #367, opened 2024-02-05
connect internal address to a vsi

olasaadi99:


pkg/vpcmodel/nodesExplainability.go
```diff
@@ -102,6 +104,21 @@ func (c *VPCConfig) getVsiNode(name string) Node {
...
 
+// given a string address in non-cidr format return the list of all network interfaces attached to it
+func (c *VPCConfig) GetNetworkInterfaces(internalAddress string) (networkInterfaceNodes []Node, err error) {
+	if err != nil {
+		return nil, err
+	}
+
+	networkInterfaceNodes = []Node{}
+	for _, node := range c.Nodes {
+		if node.Kind() == resourceTypeNetworkInterface && node.Cidr() == internalAddress {
```
adisos, line 115: @ShiriMoran do we really need to check `node.Kind()` and cannot just skip nodes that are `IsExternal()` ?

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -10,6 +10,8 @@ import (
 
 var filterLayers = [2]string{NaclLayer, SecurityGroupLayer}
 
+const resourceTypeNetworkInterface = "NetworkInterface"
```
adisos, line 13: can we avoid duplicating the const string for network interface kind? is it a more ibm-specific term to be used only by ibmvpc pkg rather than in the abstract vpcmodel pkg?

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -102,6 +104,21 @@ func (c *VPCConfig) getVsiNode(name string) Node {
 	return nil
 }
 
+// given a string address in non-cidr format return the list of all network interfaces attached to it
+func (c *VPCConfig) GetNetworkInterfaces(internalAddress string) (networkInterfaceNodes []Node, err error) {
```
adisos, line 108: how is this function going to be used by explainability?
is it also part of this PR?

pkg/vpcmodel/nodesExplainability.go
adisos: also, the name of the new function `GetNetworkInterfaces`  hints there is an entity "network interfaces" not defined in this level.. this pkg is on level of abstract nodes

pkg/vpcmodel/nodesExplainability.go
ShiriMoran: Given an internal address, explainability will be applied to any of its VSIs (could be a Cartesian product if both src and dst given as internal cidr). If there are no VSIs connected then the user will  be provided with this explanation 
Lets finalize details in https://github.com/np-guard/vpc-network-config-analyzer/issues/353

No, it will be in a separate PR

pkg/vpcmodel/nodesExplainability.go
ShiriMoran: What I had in mind was connecting specifically to network interfaces, assuming that internal nodes can represent other resources. But I'm also OK with returning all internal Nodes, and having more specific checks in other functionality, probably withing ibmvpc

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -102,6 +102,29 @@ func (c *VPCConfig) getVsiNode(name string) Node {
...
```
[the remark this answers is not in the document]
olasaadi99: > * is the input `internalAddress` only an exact ip address, or also possibly a cidr  ? the current name `internalAddress` may be misleading, as it seems the implementation supports both
@ShiriMoran asked to start by only an exact ip address.. but it is the same implementation

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -102,6 +102,27 @@ func (c *VPCConfig) getVsiNode(name string) Node {
 	return nil
 }
 
+// given a string address in cider format or non-cidr format return the list of all network interfaces attached to it
```
adisos, line 105:
```suggestion
// GetNodesWithinAddress is given a string address in CIDR format or exact IP address format, and returns the list of all internal nodes with this address
```

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -102,6 +102,27 @@ func (c *VPCConfig) getVsiNode(name string) Node {
...
+	addressIPblock = common.NewIPBlockFromCidrOrAddress(ipAddress)
+
+	networkInterfaceNodes = []Node{}
+	for _, node := range c.Nodes {
+		if networkInterfaceIPBlock, err = common.NewIPBlockFromIPAddress(node.Cidr()); err != nil {
+			return nil, err
+		}
+		contained := networkInterfaceIPBlock.ContainedIn(addressIPblock)
+		if node.IsExternal() && contained {
+			return nil, fmt.Errorf("src or dst address %v represents an external IP", ipAddress)
```
adisos, line 117:
```suggestion
			return nil, fmt.Errorf("src or dst address %s represents an external IP", ipAddress)
```

adisos (approved):

PR #371, opened 2024-02-07
extend input to VSI

olasaadi99:


pkg/vpcmodel/nodesExplainability.go
```diff
@@ -131,6 +133,20 @@ func (c *VPCConfig) GetNodesWithinAddress(ipAddress string) (networkInterfaceNod
 	return networkInterfaceNodes, nil
 }
 
+// getNodesOfVsi is given a string name or UID of VSI, and
+// returns the list of all nodes with this vsi name or UID
+func (c *VPCConfig) GetNodesOfVsi(vsi string) (vsiNodes []Node, err error) {
+	for _, nodeSet := range c.NodeSets {
+		if nodeSet.Kind() == ResourceTypeVSI && (nodeSet.Name() == vsi || nodeSet.UID() == vsi) {
```
ShiriMoran, line 140: Not exactly. What I meant was that there are at least two different `NodeSets` such that in each of them the given name equals the Name or the UID. Its not likely it will be in the same `NodeSets`

adisos: Seems the comparison is for different NodeSets

adisos, line 140: I think we can avoid checking `nodeSet.Kind() == ResourceTypeVSI` (the VSI is ibm-specific concept), and just return the NodeSet object matching the input name/UID.. 
@ShiriMoran  please confirm

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -131,6 +133,20 @@ func (c *VPCConfig) GetNodesWithinAddress(ipAddress string) (networkInterfaceNod
 	return networkInterfaceNodes, nil
 }
 
+// getNodesOfVsi is given a string name or UID of VSI, and
+// returns the list of all nodes with this vsi name or UID
```
adisos, line 137:
```suggestion
// returns the list of all nodes within this vsi
```

pkg/vpcmodel/nodesExplainability.go
ShiriMoran: @adisos `NodeSet` can also be e.g. `subnet`; if the user entered a `subnet` name (not very likely but possible) we may get unexpectable results.
Perhaps we can move the function to `ibmvpc`?

olasaadi99: I do keep the previous nodeSet..

ShiriMoran: So please in the error message notify which are the two NodeSets

ShiriMoran: But what if the given name matches exactly one NodeSets which is a subnet and not a VSI?

adisos: Maybe add a "todo" comment to handle this later.. we have another open issue (#380)  for adding abstract vpc and subnet, which could be used here..

ShiriMoran: @olasaadi99 please add todo

ShiriMoran (approved):

PR #374, opened 2024-02-07
357 explainability finalize output

ShiriMoran:


ShiriMoran:
replaced by https://github.com/np-guard/vpc-network-config-analyzer/pull/378

PR #390, opened 2024-02-15
379 connect vsi internal ip input

ShiriMoran:


pkg/ibmvpc/explainability_test.go
```diff
@@ -436,17 +453,26 @@ func TestInputValidity(t *testing.T) {
...
+	// should fail due to non existing networkInterface
+	_, err2 := vpcConfig.ExplainConnectivity(cidr1, nonExistingNif, nil)
 	fmt.Println(err2.Error())
 	if err2 == nil {
-		require.Fail(t, err1.Error())
+		require.Fail(t, err2.Error())
+	}
+	_, err3 := vpcConfig.ExplainConnectivity(cidrAll, existingNif, nil)
+	fmt.Println(err3.Error())
+	if err3 == nil {
```
adisos, line 475: each `if` here can be shortened to one line using `require.NotNil` or `require.Nil`

pkg/ibmvpc/explainability_test.go
```diff
@@ -436,17 +453,26 @@ func TestInputValidity(t *testing.T) {
...
 	}
-	_, err2 := vpcConfig.ExplainConnectivity(cidr1, nonExistingVSI, nil)
+	// should fail due to non existing networkInterface
+	_, err2 := vpcConfig.ExplainConnectivity(cidr1, nonExistingNif, nil)
 	fmt.Println(err2.Error())
 	if err2 == nil {
-		require.Fail(t, err1.Error())
+		require.Fail(t, err2.Error())
+	}
+	_, err3 := vpcConfig.ExplainConnectivity(cidrAll, existingNif, nil)
```
adisos, line 473: why should fail here?

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -230,27 +115,37 @@ func (c *VPCConfig) ExplainConnectivity(src, dst string, connQuery *common.Conne
...
 	conn *common.ConnectionSet) (rulesAndConn rulesAndConnDetails, err error) {
-	rulesAndConn = make(rulesAndConnDetails, max(len(srcNodes), len(dstNodes)))
-	i := 0
-	// either src of dst has more than one item; never both
-	// the loop is on two dimension since we do not know which, but actually we have a single dimension
+	// the size is not known in this stage due to the corner case in which we have the same node both in srcNodes and dstNodes
+	rulesAndConn = make(rulesAndConnDetails, 0)
 	for _, src := range srcNodes {
 		for _, dst := range dstNodes {
+			if src.Name() == dst.Name() {
```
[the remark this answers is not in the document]
ShiriMoran: indeed. done.

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/np-guard/vpc-network-config-analyzer/pkg/common"
+)
+
+const noValidInputErr = "does not represent a VSI, an internal interface, an internal IP with network interface or " +
+	"a valid external IP"
+
+// GetConnectionSet TODO: handle also input ICMP properties (type, code) as input args
+// translates explanation args to a connection set
+func (e *ExplanationArgs) GetConnectionSet() *common.ConnectionSet {
```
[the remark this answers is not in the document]
ShiriMoran: right. done.

pkg/ibmvpc/explainability_test.go
ShiriMoran: // should fail due to a cidr containing both public internet and internal address
added to the code
(funny I forget to mention it)

pkg/ibmvpc/explainability_test.go
ShiriMoran: fixed (also fixed corresponding error msg)

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,212 @@
...
+	if inputIPBlock == nil { // 1. string cidr does not represent a legal cidr
+		return nil, false, nil
+	}
+	// 2.
+	_, publicInternet, err1 := getPublicInternetIPblocksList()
+	if err1 != nil {
+		return nil, false, err1
+	}
+	isExternal := !inputIPBlock.Intersection(publicInternet).Empty()
+	isInternal := !inputIPBlock.ContainedIn(publicInternet)
```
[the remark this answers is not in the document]
ShiriMoran: currently we are not handling these addresses - which are not within VPC and also not public internet. I guess in this stage - in which we are not yet supporting tgw - we should exit with error if such an address is reference (even as part of a larger address - to make sure we are not misleading). Agreed @adisos?

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,212 @@
...
+// (and the nodes are its network interfaces). Specifically:
+//  1. If it does not present a cidr or IP, returns nil and false
+//  2. If it represents a cidr which is both internal and external, returns an error
+//  3. If it presents an external address, returns external addresses nodes and false
+//  4. If it presents an internal address, return connected network interfaces if any and true,
+//     error otherwise
+func (c *VPCConfig) getNodesFromAddress(ipOrCidr string) (nodes []Node, internalIP bool, err error) {
+	inputIPBlock := common.NewIPBlockFromCidrOrAddress(ipOrCidr)
+	if inputIPBlock == nil { // 1. string cidr does not represent a legal cidr
+		return nil, false, nil
```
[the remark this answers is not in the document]
ShiriMoran: This case will result in an error; however, the error is broader than just the "`IP/cidr context`". This does not return an error for the same reason `getNodesOfVsi` and `getNetworkInterfaceNodes` do not return an error if no match. 
If the `src/dst` input string does not match anything then an error of no match will be given.

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,212 @@
...
+		connection.AddTCPorUDPConn(common.ProtocolStr(e.protocol), e.srcMinPort,
+			e.srcMaxPort, e.dstMinPort, e.dstMaxPort)
+	}
+
+	return connection
+}
+
+// given src and dst input finds the []nodes they represent
+// src/dst may refer to:
+// 1. NetworkInterface by name
```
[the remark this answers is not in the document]
ShiriMoran: Just to make sure we are not the same page:
the former comment, which I agree to, requires change in the code.
the latter required only writing the correct input help  and README 
In the latter I would leave  the name - `UID` - to be consistent with the name used in the relevant interface.

BTW, does the `UID` hold the `CRN` and not the actual `UID`? @adisos @zivnevo

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,212 @@
...
+// 2. VSI by UID or name; in this case we consider the network interfaces of the VSI
+// 3. Internal IP address; in this case we consider the vsis in that address range
+// 4. external IP address
```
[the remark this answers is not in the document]
ShiriMoran: Lets agree on how to resolve https://github.com/np-guard/vpc-network-config-analyzer/pull/390#discussion_r1491011243 first

pkg/vpcmodel/explainabilityInput.go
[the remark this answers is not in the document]
ShiriMoran: as agreed, I removed the option of using our internal presentation of network interface as src/dst input

pkg/vpcmodel/explainabilityInput.go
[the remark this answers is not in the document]
ShiriMoran: as agreed, added error message in such cases

pkg/vpcmodel/explainabilityInput.go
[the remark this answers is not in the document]
ShiriMoran: done

pkg/vpcmodel/explainabilityInput.go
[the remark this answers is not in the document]
adisos: This type's methods (`Src()` and `Dst() ) should also move to this file.

cmd/analyzer/parse_args.go
```diff
@@ -186,8 +186,9 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
 	args.VPC = flagset.String(VPC, "", "CRN of the VPC to analyze")
 	args.Debug = flagset.Bool(Debug, false, "Run in debug mode")
 	args.Version = flagset.Bool(Version, false, "Prints the release version number")
-	args.ESrc = flagset.String(ESrc, "", "Source name for network_interface or an external IP to be explained")
-	args.EDst = flagset.String(EDst, "", "Destination name for network_interface or an external IP to be explained")
+	args.ESrc = flagset.String(ESrc, "", "Source name for explanation; "+
```
adisos, line 189:
```suggestion
	args.ESrc = flagset.String(ESrc, "", "Source endpoint for explanation; "+
```

cmd/analyzer/parse_args.go
```diff
@@ -186,8 +186,9 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
...
+		"can be specified as a vsi name or an internal or external address (IP/cidr)")
```
adisos, line 190:
```suggestion
		"can be specified as a VSI name/CRN or an internal/external IP-address/CIDR")
```

pkg/ibmvpc/explainability_test.go
```diff
@@ -34,77 +34,77 @@ var explainTests = []*vpcGeneralTest{
 	{
 		name:        "VsiToVsi1",
 		inputConfig: "sg_testing1_new",
-		ESrc:        "vsi2-ky[10.240.20.4]",
-		EDst:        "vsi3b-ky[10.240.30.4]",
+		ESrc:        "vsi2-ky",
```
adisos, line 37: is there a test with src or dst as CRN of a VSI?

pkg/vpcmodel/explainabilityInput.go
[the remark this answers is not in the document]
ShiriMoran: right. done

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	if nodeSetWithVsi == nil {
+		return nil, nil
+	}
+	return nodeSetWithVsi.Nodes(), nil
+}
+
+// getNodesFromAddress gets a string that should present a cidr or IP
+// and returns the corresponding node(s) and a bool which is true iff cidrOrName is an internal address
```
adisos, line 121:
```suggestion
// and returns the corresponding node(s) and a bool which is true iff ipOrCidr is an internal address
```

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -0,0 +1,226 @@
...
+		return nodes, false, err
+		// internal address
+	} else if isInternal {
+		// 4.
+		vpcAP := c.VPC.AddressRange()
+		if !inputIPBlock.ContainedIn(vpcAP) {
+			return nil, false, fmt.Errorf("internal address %s not within vpc's address prefix %s",
+				inputIPBlock.ToIPRanges(), vpcAP.ToIPRanges())
+		}
+		// 5.
```
adisos, line 159: `VPC.AddressRange()` returns the ranges for VPC's subnets and not the address prefix.

ShiriMoran: as agreed, we'll change the relevant error msg add a todo (better have address prefix, but at the moment not worth the effort)

cmd/analyzer/parse_args.go
```diff
@@ -186,8 +186,9 @@ func ParseInArgs(cmdlineArgs []string) (*InArgs, error) {
 	args.VPC = flagset.String(VPC, "", "CRN of the VPC to analyze")
 	args.Debug = flagset.Bool(Debug, false, "Run in debug mode")
 	args.Version = flagset.Bool(Version, false, "Prints the release version number")
-	args.ESrc = flagset.String(ESrc, "", "Source name for network_interface or an external IP to be explained")
-	args.EDst = flagset.String(EDst, "", "Destination name for network_interface or an external IP to be explained")
+	args.ESrc = flagset.String(ESrc, "", "Source endpoint for explanation; "+
+		"can be specified as a VSI name/CRN or an internal/external IP-address/CIDR")
+	args.EDst = flagset.String(EDst, "", "Destination name for explanation; same format as src")
```
adisos, line 191:
```suggestion
	args.EDst = flagset.String(EDst, "", "Destination endpoint for explanation; same format as src")
```

README.md
```diff
@@ -20,7 +20,7 @@ Usage of vpc-network-config-analyzer:
   -debug
         Run in debug mode
   -dst string
-        Destination name for network_interface or an external IP to be explained
+        Destination name for explaination; can be specified as a vsi name or an internal or external address (IP/cidr)
```
adisos, line 23: update based on additional changes in `parse_args.go`

README.md
```diff
@@ -36,7 +36,7 @@ Usage of vpc-network-config-analyzer:
   -protocol string
         Protocol for connection description
   -src string
-        Source name for network_interface or an external IP to be explained
+        Source name for explaination; can be specified as a vsi name or an internal or external address (IP/cidr)
```
adisos, line 39: here too

pkg/ibmvpc/explainability_test.go
ShiriMoran: No, I was not sure how useful the crn will be.
The point here is that VSI name may not be unique - so why use crn for UID and not the actual id?
Plus, currently our "crn" is not a real crn thus a test might not be useful, as it will not test a real crn which has certain non standard characters.

Anyways, I've added a test. Should I open an issue (for using id instead of crn)?

adisos: The UID for any resource in the model is the CRN, except for few resources that do not have CRN, for which ID is used. 
(both CRN and ID are unique). 
It was decided to use CRN when possible when started modeling. (`parser.go` assigns those values).
I don't think there is an issue in testing a fake crn, just need to make sure that it matches the right field from the config object.

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -100,6 +100,7 @@ func (c *VPCConfig) getNodesFromInputString(cidrOrName string) (nodes []Node, in
 func (c *VPCConfig) getNodesOfVsi(vsi string) ([]Node, error) {
 	var nodeSetWithVsi NodeSet
 	for _, nodeSet := range c.NodeSets {
+		fmt.Printf("name: %v uid: %v\n", nodeSet.Name(), nodeSet.UID())
```
adisos, line 103: remove this `Printf`

pkg/vpcmodel/textOutput.go
```diff
@@ -40,13 +41,28 @@ func headerOfAnalyzedVPC(uc OutputUseCase, vpcName, vpc2Name string, c1 *VPCConf
...
 		}
-		return fmt.Sprintf("Connectivity explanation%s between %s and "+
-			"%s\n===================================================================================================\n\n",
-			connStr, explanation.src, explanation.dst), nil
+		srcNetworkInterfaces := explainNetworkInterfaces(explanation.srcNetworkInterfacesFromIP)
+		dstNetworkInterfaces := explainNetworkInterfaces(explanation.dstNetworkInterfacesFromIP)
+		header1 := fmt.Sprintf("Connectivity explanation%s between %s%s and %s%s",
+			connStr, explanation.src, srcNetworkInterfaces, explanation.dst, dstNetworkInterfaces)
+		header2 := strings.Repeat("=", len(header1))
+		return header1 + "\n" + header2 + "\n\n", nil
```
adisos, line 49: this should be a method of `Explanation` ?

pkg/ibmvpc/explainability_test.go
ShiriMoran: I've modified the "crn" test to have an exact "crn" format, which includes (of course) changing the relevant test file.

However, my point is that it is very unlikely anyone will actually use the `crn` as input. This is not the case for the `id`.  Why not use the id as UID?

pkg/vpcmodel/explainabilityInput.go
ShiriMoran: oops
removed

README.md
ShiriMoran: done. I've also changed dst to no reference src

README.md
ShiriMoran: done

pkg/ibmvpc/explainability_test.go
adisos: It was decided to use crn and not id as Karen suggested.
@kyorav what do you think?

pkg/vpcmodel/textOutput.go
ShiriMoran: I don't think that "pure" printing related functions should reside in `Explanation`.

Do you think `listNetworkInterfaces` (perhaps without the "(" + ") ) can be useful in other cases? perhaps this function can be moved to a common file? Or maybe to Explanation?

adisos: I would move this code (under `case Explain`) to a function such as `explainabilityReportHeader(explanation *Explanation)`, and maybe have both this new function and `explainNetworkInterfaces` in a file related to explainability.

ShiriMoran: I've added a file for printing related to explanation, and moved other relevant functions

pkg/vpcmodel/textOutput.go
```diff
@@ -36,13 +36,8 @@ func headerOfAnalyzedVPC(uc OutputUseCase, vpcName, vpc2Name string, c1 *VPCConf
...
 	case Explain:
-		connStr := ""
-		if explanation.connQuery != nil {
-			connStr = " for " + explanation.connQuery.String()
-		}
-		return fmt.Sprintf("Connectivity explanation%s between %s and "+
-			"%s\n===================================================================================================\n\n",
-			connStr, explanation.src, explanation.dst), nil
+		header := explainHeader(explanation)
+		return header, nil
```
adisos, line 40:
```suggestion
		return explainHeader(explanation), nil
```

adisos (commented):
just another small typo

adisos (approved):

Issue #344, opened 2024-01-28
Investigate an ignored error of *nacl 


ShiriMoran:
In:
`func (n *NACL) AllowedConnectivity(src, dst vpcmodel.Node, isIngress bool)`
error returned from `common.IsAddressInSubnet` is ignore.
Trying not to ignore the error results in many tests failing.
This has to be investigated and fixed

adisos:
The issue was that `common.IsAddressInSubnet` handled only IP Address, but was called also for external nodes with CIDR instead of IP Address (though should not have been called for external nodes at all). 
Fixed in PR #394 .

ShiriMoran:
`The` reason ignoring the error did not cause any errors was that the ignores case were not in any `subnet` and so were treated correctly.

Issue #388, opened 2024-02-14
Node interface issue


adisos:
`Node.Cidr()` returns cidr for external nodes, but IP address for network interfaces.

adisos:
PR #394  - renamed this interface method to `CidrOrAddress`, so that it returns cidr for external nodes and address for internal nodes.

PR #394, opened 2024-02-18
Ipblock model issues

adisos:
- Better error handling for `NewIPBlockFrom...` functions
- Add IPBlock fields for structs implementing nodes, and new interface method for Node: `IPBlock()`
- issue #388 - Rename `Cidr()` method for `Node` to `CidrOrAddress()`  (returns cidr for external node and address for internal node)
- Resolves issue #344  
- Few more simplifications


pkg/common/ipBlock.go
```diff
@@ -186,32 +186,44 @@ func addIntervalToList(ipbNew *IPBlock, ipbList []*IPBlock) []*IPBlock {
...
 	if err != nil {
-		return nil
+		return nil, err
 	}
-	return res
+	return res, nil
+}
+
+// PairCIDRsToIPBlocks returns two IPBlock objects from two input CIDR strings
+func PairCIDRsToIPBlocks(cidr1, cidr2 string) (ipb1, ipb2 *IPBlock, err error) {
```
ShiriMoran, line 199: This function is called twice from `ibmVPC` - wouldn't it make more sense to put it there?

adisos: I think it is general functionality for `IPBlock`, might be used by other packages in the future..

ShiriMoran (commented):
Comments for the first commit.

pkg/vpcmodel/abstractVPC.go
```diff
@@ -66,9 +66,14 @@ const (
 
 type Node interface {
 	VPCResourceIntf
-	Cidr() string
+	// CidrOrAddress returns the string of the Node's ip-address (for internal node) or cidr ) (for external node)
+	CidrOrAddress() string
```
ShiriMoran, line 70: This is much better than before. Still, I would minimize the usage of this ambiguity function to the very minimal - e.g. printing functions.
Perhaps restore the function `Cidr()` and have it return an empty string for internal `Nodes`? And then use each `Cidr(), IPBlock()` in the relevant context and use `CidrOrAddress()` only when it can indeed be both

pkg/ibmvpc/parser.go
```diff
@@ -426,7 +426,7 @@ func getFipConfig(
 				// a node captured by a fip should not be captured by a pgw
 				for _, nodeWithFip := range srcNodes {
 					if vpcmodel.HasNode(pgw.Sources(), nodeWithFip) {
-						pgw.src = getCertainNodes(pgw.Sources(), func(n vpcmodel.Node) bool { return n.Cidr() != nodeWithFip.Cidr() })
+						pgw.src = getCertainNodes(pgw.Sources(), func(n vpcmodel.Node) bool { return n.CidrOrAddress() != nodeWithFip.CidrOrAddress() })
```
ShiriMoran, line 429: Can the returned value here be indeed `cidr` or `address`?
(See general comment in this regard in `abstractVPC`)

pkg/ibmvpc/parser.go
```diff
@@ -1030,7 +1030,7 @@ func printConfig(c *vpcmodel.VPCConfig) {
 		if n.IsExternal() {
 			continue
 		}
-		fmt.Println(strings.Join([]string{n.Kind(), n.Cidr(), n.Name(), n.UID()}, separator))
+		fmt.Println(strings.Join([]string{n.Kind(), n.CidrOrAddress(), n.Name(), n.UID()}, separator))
```
ShiriMoran, line 1043: same comment...

pkg/ibmvpc/vpc.go
```diff
@@ -45,7 +45,7 @@ type ReservedIP struct {
 	vpe     string
 }
 
-func (r *ReservedIP) Cidr() string {
+func (r *ReservedIP) CidrOrAddress() string {
```
ShiriMoran, line 48: same comment

pkg/ibmvpc/vpc.go
```diff
@@ -76,7 +76,7 @@ type NetworkInterface struct {
 	ipblock *common.IPBlock
 }
 
-func (ni *NetworkInterface) Cidr() string {
+func (ni *NetworkInterface) CidrOrAddress() string {
```
ShiriMoran, line 77: same comment, also in other relevant places (I'll stop repeating it)

pkg/ibmvpc/vpc.go
```diff
@@ -417,65 +417,65 @@ func (n *NACL) GeneralConnectivityPerSubnet(subnet *Subnet) string {
...
+		return concreteNode.subnet, nil
 	case *IKSNode:
-		return t.subnet.cidr, t.CidrOrAddress(), t.subnet, nil
+		return concreteNode.subnet, nil
 	case *ReservedIP:
-		return t.subnet.cidr, t.CidrOrAddress(), t.subnet, nil
+		return concreteNode.subnet, nil
 	default:
-		return "", "", nil, fmt.Errorf("cannot get cidr for node: %+v", n)
+		return nil, fmt.Errorf("cannot get cidr for node: %+v", node)
```
ShiriMoran, line 425: Correct the error message - the failure here is of getting `subnet` for node (when reviewing each commit separately you can't suggest changes)

pkg/ibmvpc/vpc.go
```diff
@@ -417,65 +417,65 @@ func (n *NACL) GeneralConnectivityPerSubnet(subnet *Subnet) string {
...
 		return nil, err
 	}
 	// check if the subnet of the given node is affected by this nacl
-	if _, ok := n.subnets[subnetCidr]; !ok {
+	if _, ok := n.subnets[subnet.cidr]; !ok {
 		return vpcmodel.NoConns(), nil // not affected by current nacl
 	}
 	// TODO: differentiate between "has no effect" vs "affects with allow-all / allow-none "
-	if targetNode.IPBlock().ContainedIn(subnet.ipblock) {
+	if targetNode.IsInternal() && targetNode.IPBlock().ContainedIn(subnet.ipblock) {
```
ShiriMoran, line 450: This is the place of the #344 bug - but after the changes, why is the check `targetNode.IsInternal()` needed?
`IPBlock` is well defined for external addresses

ShiriMoran (commented):
comments for the last 3 commits from Saturday

pkg/ibmvpc/parser.go
```diff
@@ -776,7 +787,25 @@ func getSubnetByIPAddress(address string, c *vpcmodel.VPCConfig) (subnet *Subnet
...
+	return nil, fmt.Errorf("could not find matching subnet for address %s", addressIPblock.ToIPAdressString())
+}
+
+func setNodeIPBlock(node vpcmodel.Node, address string) error {
+	addressIPBlock, err := common.NewIPBlockFromIPAddress(address)
+	if err != nil {
+		return err
+	}
+	switch nodeConcrete := node.(type) {
+	case *ReservedIP:
```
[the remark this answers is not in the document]
adisos: because each case converts to a different type, and should access the field `ipblock` for each type separately (the input object is of interface `Node` that does not have this functionality).

pkg/ibmvpc/connectivityAnalysis_test.go
```diff
@@ -379,3 +381,8 @@ func TestAnalyzeConnectivity(t *testing.T) {
 	fmt.Println(connectivityStr)
 	fmt.Println("done")
 }
+
+func newIPBlockFromValidatedCIDROrAddress(cidr string) *common.IPBlock {
```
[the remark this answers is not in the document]
adisos: It is indeed for testing, the validation is not required, since the calls are from hard-coded cidrs on each test, which are assumed to be valid.

pkg/common/ipBlock.go
```diff
@@ -186,32 +186,32 @@ func addIntervalToList(ipbNew *IPBlock, ipbList []*IPBlock) []*IPBlock {
 	return ipbList
 }
 
-func NewIPBlockFromCidr(cidr string) *IPBlock {
+func NewIPBlockFromCidr(cidr string) (*IPBlock, error) {
 	res, err := NewIPBlock(cidr, []string{})
```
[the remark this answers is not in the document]
adisos: Right, changed.
I would keep both functions `NewIPBlockFromCidr` and `NewIPBlock` with the exception list. 
This `IPBlock` pkg is going to move to another repo, that may later be consumed by netpol-analyzer, where the function with exception list is relevant.

pkg/ibmvpc/nacl_analysis.go
```diff
@@ -453,28 +453,37 @@ func (na *NACLAnalyzer) GeneralConnectivityPerSubnet(subnetCidr string) (
...
 	}
-	targetIPblock = common.NewIPBlockFromCidrOrAddress(target)
-	inSubnetIPblock = common.NewIPBlockFromCidrOrAddress(inSubentCidr)
-	return analyzedConns, targetIPblock, inSubnetIPblock
+	targetIPblock, err1 := common.NewIPBlockFromCidrOrAddress(target)
+	inSubnetIPblock, err2 := common.NewIPBlockFromCidrOrAddress(inSubentCidr)
+	if err1 != nil && err2 != nil {
+		return nil, nil, nil, errors.Join(err1, err2)
+	}
+	return analyzedConns, targetIPblock, inSubnetIPblock, nil
```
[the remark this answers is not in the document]
adisos: This is about outdated code

pkg/ibmvpc/parser.go
```diff
@@ -873,6 +893,11 @@ func getIKSnodesConfig(res map[string]*vpcmodel.VPCConfig,
 			address: *iksNodeNetIntfObj.IpAddress,
 			subnet:  subnet,
 		}
+		addressIPBlock, err := common.NewIPBlockFromIPAddress(nodeObject.address)
+		if err != nil {
+			return err
+		}
+		nodeObject.ipblock = addressIPBlock
```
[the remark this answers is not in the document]
adisos: Currently not.
The config object is currently assumed to be valid (there is an issue to add validation of config objects while parsing, maybe this is also related to this issue).

pkg/ibmvpc/vpc.go
adisos: done

pkg/ibmvpc/vpc.go
adisos: Just to save a call to `ContainedIn` when not necessary

pkg/ibmvpc/parser.go
adisos: This will be changed in a follow up PR (#396 ), the comparison will be based on UID instead of address.

pkg/vpcmodel/abstractVPC.go
adisos: Indeed I tried minimize the usage of this function.

I think it might be confusing/error-prone to have a `Cidr()` function that sometimes returns an empty string. 
The `IPBlock()` function is not a replacement for `CidrOrAddress()`, it does not return an IP address string, just an IPBlock object.

pkg/ibmvpc/parser.go
adisos: This is just a printing function, so I don't find it too confusing here. 
From the function documentation it is known that it returns an address for internal node, which is the case here.

pkg/ibmvpc/connectivityAnalysis_test.go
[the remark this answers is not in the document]
ShiriMoran: So perhaps rename the func; now it is misleading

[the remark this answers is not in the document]
adisos: renamed to `newIPBlockFromCIDROrAddressWithoutValidation`

pkg/vpcmodel/abstractVPC.go
ShiriMoran: To my understanding, this function is used when we need really need is `Cidr`. Personally I find it confusing and even error prune.
We are calling this function almost 50 times, which is too much, I think. 
Seems there is something we are not doing right here. Should we open a follow up issue? at the very minimal I would leave a README

pkg/ibmvpc/parser.go
ShiriMoran: Also in other places we are calling the function CidrOrAddress() but expect only Cidr?

pkg/ibmvpc/parser.go
ShiriMoran: yes, this is an example of a proper use of the function

pkg/ibmvpc/vpc.go
ShiriMoran: perhaps add a comment, to avoid confusing

pkg/vpcmodel/abstractVPC.go
adisos: > this function is used when we need really need is Cidr

Not necessarily, for internal nodes we sometimes really need an address.

> We are calling this function almost 50 times

Less. these are the places where this is called (after the follow up PR that also removes some calls):
```
pkg/ibmvpc/nacl_analysis.go:    return nil, fmt.Errorf(notFoundMsg, isIngress, targetNode.CidrOrAddress(), subnet.cidr, nodeInSubnet.CidrOrAddress())
pkg/ibmvpc/nacl_analysis.go:    return nil, nil, fmt.Errorf(notFoundMsg, isIngress, targetNode.CidrOrAddress(), subnet.cidr, nodeInSubnet.CidrOrAddress())
pkg/ibmvpc/parser.go:           fmt.Println(strings.Join([]string{n.Kind(), n.CidrOrAddress(), n.Name(), n.UID()}, separator))
pkg/ibmvpc/vpc.go:      return member.CidrOrAddress(), target.IPBlock()
pkg/vpcmodel/drawioGenerator.go:                tooltip = append(tooltip, n.(*ExternalNetwork).CidrOrAddress())
pkg/vpcmodel/externalNetwork.go:                ipList[i] = n.CidrOrAddress()
pkg/vpcmodel/grouping.go:               cidrList[i] = n.CidrOrAddress()
pkg/vpcmodel/nodesConnectivity.go:                      srcName := srcNode.CidrOrAddress()
pkg/vpcmodel/nodesConnectivity.go:                      dstName := dstNode.CidrOrAddress()
pkg/vpcmodel/nodesConnectivity.go:                      strList = append(strList, getConnectionStr(peerNode.CidrOrAddress(), node.CidrOrAddress(), conn.String(), " [inbound]"))
pkg/vpcmodel/nodesConnectivity.go:                      strList = append(strList, getConnectionStr(node.CidrOrAddress(), peerNode.CidrOrAddress(), conn.String(), " [outbound]"))
pkg/vpcmodel/nodesConnectivity.go:                      strList = append(strList, getConnectionStr(src.(Node).CidrOrAddress(), dst.(Node).CidrOrAddress(), conns.String(), ""))
pkg/vpcmodel/semanticDiff.go:                   nodeSameCidr := findNodeWithCidr(other.Nodes, ep.(Node).CidrOrAddress())
pkg/vpcmodel/semanticDiff.go:           if node.CidrOrAddress() == cidr {
```

> Should we open a follow up issue

Yes, maybe have a follow up issue, to see if/how this can be improved.

pkg/ibmvpc/vpc.go
adisos: added comment

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -164,17 +164,17 @@ func (c *VPCConfig) GetNodesOfVsi(vsi string) ([]Node, error) {
 //  2. Calculate from N and the cidr block, disjoint IP blocks
 //  3. Return the nodes created from each block from 2 contained in the input cidr
 func (c *VPCConfig) getCidrExternalNodes(cidr string) (cidrNodes []Node, err error) {
-	cidrsIPBlock := common.NewIPBlockFromCidrOrAddress(cidr)
-	if cidrsIPBlock == nil { // string cidr does not represent a legal cidr
-		return nil, nil
+	cidrsIPBlock, err := common.NewIPBlockFromCidrOrAddress(cidr)
+	if err != nil { // string cidr does not represent a legal cidr
+		return nil, err
```
adisos, line 169: @ShiriMoran  should this be changed? (avoid returning the error returned from `NewIPBlockFromCidrOrAddress`, and just return a `nil` object) ?

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -115,15 +115,15 @@ func (c *VPCConfig) getVsiNode(name string) Node {
 // GetNodesWithinAddress is given a string address in CIDR format or exact IP address format, and
 // returns the list of all internal nodes with this address
 func (c *VPCConfig) GetNodesWithinAddress(ipAddress string) (networkInterfaceNodes []Node, err error) {
-	var addressIPblock, networkInterfaceIPBlock *common.IPBlock
-	addressIPblock = common.NewIPBlockFromCidrOrAddress(ipAddress)
+	var addressIPblock *common.IPBlock
+	addressIPblock, err = common.NewIPBlockFromCidrOrAddress(ipAddress)
+	if err != nil {
+		return nil, err
+	}
```
adisos, line 122: same question here

pkg/vpcmodel/nodesExplainability.go
```diff
@@ -449,6 +449,7 @@ func (c *VPCConfig) processInput(srcName, dstName string) (srcNodes, dstNodes []
 	if err != nil {
 		return nil, nil, err
 	}
+	// TODO: next 'if' seems to be redundant now
```
adisos, line 452: I assume this comment can be deleted if we do remove the errors returned from IPBlock generation earlier, and then checking that `len(srcNodes)` is not redundant, right?

pkg/ibmvpc/parser.go
adisos: Yes, for example in `isEntirePublicInternetRange` within `externalNetwork.go`. The input is a list of nodes, but all are expected to be external nodes. 
One option is to convert `Node` to `ExternalNetwork`, and use its `CidrStr` field.
If the type `groupedExternalNodes` was defined as `[]ExternalNetwork` instead of `[]Node`, then conversion would not be required in this case.

pkg/vpcmodel/abstractVPC.go
ShiriMoran: opened #397 and assigned both of us for now

pkg/vpcmodel/nodesExplainability.go
ShiriMoran: yes, because we don't know whether the user meant to insert an `IPBlock/Cidr`. If this failed all we know is that what he inserted does not match any legal src/dst. For example, we will get there if instead of vsi1-ky, VSI1-KY was inserted (we decided not be oblivious to cases in this stage) and the error message from `NewIPBlockFromCidrOrAddress` will be misleading. We could use some string matching to "guess" what the user meant to do, but this is surely not worth the effort for the 1st release

pkg/vpcmodel/nodesExplainability.go
ShiriMoran: yes, because we don't know whether the user meant to insert an `IPBlock/Cidr`. If this failed all we know is that what he inserted does not match any legal src/dst. For example, we will get there if instead of vsi1-ky, VSI1-KY was inserted (we decided not be oblivious to cases in this stage) and the error message from `NewIPBlockFromCidrOrAddress` will be misleading. We could use some string matching to "guess" what the user meant to do, but this is surely not worth the effort for the 1st release

ShiriMoran: this function is no longer used. Removed

pkg/vpcmodel/nodesExplainability.go
ShiriMoran: This is part was rewritten in my now merged PR. This code is no longer relevant

ShiriMoran (dismissed):

pkg/vpcmodel/nodesExplainability.go
adisos: The linter had this issue 
```
pkg\vpcmodel\explainabilityInput.go:136:3: error is not nil (line 133) but it returns nil 
```
so I've made a few more changes (I think it made sense, found it also confusing to return `nil` when `err` was not `nil`), so now this function is called only after validating that the input represents a valid IP address or CIDR.

zivnevo (changes requested):

pkg/vpcmodel/externalNetwork.go
```diff
@@ -97,17 +105,28 @@ func newExternalNode(isPublicInternet bool, ipb *common.IPBlock) (Node, error) {
...
+	cidrIPBlodk, err := common.NewIPBlockFromCidr(cidr)
+	if err != nil {
+		return nil, err
+	}
 	return &ExternalNetwork{
 		ResourceType:     publicInternetNodeName,
 		CidrStr:          cidr,
-		isPublicInternet: isPublicInternet}, nil
+		isPublicInternet: isPublicInternet,
+		ipblock:          cidrIPBlodk}, nil
```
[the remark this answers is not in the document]
adisos: right, changed

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -117,53 +121,44 @@ func (c *VPCConfig) getNodesOfVsi(vsi string) ([]Node, error) {
...
-	if inputIPBlock == nil { // 1. string cidr does not represent a legal cidr
-		return nil, false, nil
-	}
-	// 2.
 	_, publicInternet, err1 := getPublicInternetIPblocksList()
 	if err1 != nil {
 		return nil, false, err1
 	}
-	// 2.
+	// 1.
```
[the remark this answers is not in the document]
adisos: changed

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -117,53 +121,44 @@ func (c *VPCConfig) getNodesOfVsi(vsi string) ([]Node, error) {
...
 	isExternal := !inputIPBlock.Intersection(publicInternet).Empty()
 	isInternal := !inputIPBlock.ContainedIn(publicInternet)
 	if isInternal && isExternal {
 		return nil, false, fmt.Errorf("%s contains external and internal addresses which is not supported. "+
 			"src, dst should be external *or* internal address", ipOrCidr)
 	}
-	// 3.
-	if isExternal { // 3.
+	// 2.
+	if isExternal { // 2.
```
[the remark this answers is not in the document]
adisos: changed

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -117,53 +121,44 @@ func (c *VPCConfig) getNodesOfVsi(vsi string) ([]Node, error) {
...
 		}
-		// 5.
-		networkInterfaces, err2 := c.getNodesWithinInternalAddress(inputIPBlock)
-		if err2 != nil {
-			return nil, false, err2
-		}
+		// 4.
+		networkInterfaces := c.getNodesWithinInternalAddress(inputIPBlock)
 		// a given internal address should have vsi connected to it
 		if networkInterfaces == nil {
```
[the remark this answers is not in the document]
adisos: changed

zivnevo (approved):

PR #411, opened 2024-02-21
Layout tgw

haim-kermany:
1. create a specific layout for tgw, trying to optimize the location of the gtws according to the src and dst of the lines that route through the gtw
2. postponed the calculation of the routing direction to after the layout (originally it was done when creating the connection)


pkg/drawio/layout.go
```diff
@@ -593,6 +594,106 @@ func (ly *layoutS) setIconsLocationsOnTop(square SquareTreeNodeInterface) {
...
+		if col.width() >= iconSpace {
+			allCols[ci] = true
+			for _, tgw := range tgws {
+				if ci >= tgwMinCol[tgw] && ci <= tgwMaxCol[tgw] {
+					tgwOptionalCols[tgw] = append(tgwOptionalCols[tgw], ci)
+				}
+			}
+		}
+	}
+	return tgwOptionalCols, allCols
```
zivnevo, line 694: Can we be sure here that `len(tgwOptionalCols[tgw]) > 0` for every `tgw`?

zivnevo (changes requested):

pkg/drawio/layout.go
haim-kermany: I think can be sure, the minCol and maxCol are cols a src/dst, so we their width put them in tgwOptionalCols.
Anyway, made a fix to be on the safe side

pkg/drawio/layout.go
```diff
@@ -593,6 +594,106 @@ func (ly *layoutS) setIconsLocationsOnTop(square SquareTreeNodeInterface) {
...
+		var bestColAvailable int
+		bestDistance := cloud.Location().lastCol.index
+		tgwOptCol := tgwOptionalCols[tgw][0] + tgwOptionalCols[tgw][len(tgwOptionalCols[tgw])-1]/2
+		for col := range availableCols {
+			if abs(col-tgwOptCol) < bestDistance {
+				bestColAvailable = col
+				bestDistance = abs(col - tgwOptCol)
+			}
+		}
+		tgw.setLocation(newCellLocation(cloud.Location().firstRow, ly.matrix.cols[bestColAvailable]))
```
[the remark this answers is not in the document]
haim-kermany: yes,  its a bug, fixed

zivnevo (approved):

PR #424, opened 2024-02-26
340 dont group subnets diff vpc

ShiriMoran:
In subnets grouping - do not group subnets from different VPCs.
Only subnets the belong to the same VPC can be grouped.


pkg/vpcmodel/grouping.go
```diff
@@ -219,25 +219,41 @@ func (g *groupingConnections) addPublicConnectivity(ep EndpointElem, commonProps
 	(*g)[ep][connKey].appendNode(targetNode)
 }
 
-// vsiGroupingBySubnets returns a slice of EndpointElem objects, by grouping set of elements that
+// vsiOrSubnetsGroupingBySubnetsOrVsis returns a slice of EndpointElem objects, by grouping set of elements that
+// 1. If the grouped elements are vsis then the elements
 // represent network interface nodes from the same subnet into a single groupedNetworkInterfaces object
```
adisos, line 224: > into a single groupedNetworkInterfaces object

what is `groupedNetworkInterfaces ` object? should update the comment here

pkg/vpcmodel/grouping_test.go
```diff
@@ -390,40 +390,3 @@ func TestConfigSelfLoopCliqueLace(t *testing.T) {
...
-	res1.AllowedConnsCombined.updateAllowedConnsMap(res.NodeSets[0], res.NodeSets[2], common.NewConnectionSet(true))
-	res1.AllowedConnsCombined.updateAllowedConnsMap(res.NodeSets[1], res.NodeSets[0], common.NewConnectionSet(true))
-	res1.AllowedConnsCombined.updateAllowedConnsMap(res.NodeSets[1], res.NodeSets[2], common.NewConnectionSet(true))
-	res1.AllowedConnsCombined.updateAllowedConnsMap(res.NodeSets[2], res.NodeSets[0], common.NewConnectionSet(true))
-	res1.AllowedConnsCombined.updateAllowedConnsMap(res.NodeSets[2], res.NodeSets[1], common.NewConnectionSet(true))
-
-	return res, res1
-}
-
-func TestSubnetSelfLoop(t *testing.T) {
```
adisos, line 415: why is this test removed?

ShiriMoran: because now its too much trouble keeping it - the subnet needs an attached VPC with an UID - and we have this test, of clique subnets, in the testing files

pkg/vpcmodel/grouping.go
ShiriMoran: it was a typo (not from this PR), fixed

pkg/vpcmodel/grouping.go
```diff
@@ -219,25 +219,41 @@ func (g *groupingConnections) addPublicConnectivity(ep EndpointElem, commonProps
...
-func vsiGroupingBySubnets(groupedConnLines *GroupConnLines,
-	elemsList []EndpointElem, c *VPCConfig) []EndpointElem {
+// 2. If the grouped elements are vis then the elements represent subnets from the same VPC
+// this is automatic unless VPCConfig is result of a dummy vpc built for tgw, namely IsMultipleVPCsConfig = true
+func vsiOrSubnetsGroupingBySubnetsOrVsis(groupedConnLines *GroupConnLines,
```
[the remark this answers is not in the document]
ShiriMoran: comment modified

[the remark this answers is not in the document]
adisos: comment is still not accurate "map from subnet's/vsi's UID to its nodes/nodeSets from the input"
if it's a vsi, then the map key is ` n.Subnet().UID()` => the uid of the node's subnet 
if it's a subnet, then the map key is `n.VPC().UID() ` => the uid of the subnet's vpc

pkg/vpcmodel/grouping.go
```diff
@@ -219,25 +219,41 @@ func (g *groupingConnections) addPublicConnectivity(ep EndpointElem, commonProps
...
+	elemsList []EndpointElem, c *VPCConfig, groupVSI bool) []EndpointElem {
 	res := []EndpointElem{}
-	subnetNameToNodes := map[string][]EndpointElem{} // map from subnet name to its nodes from the input
+	subnetOrVSIToNodesOrNodeSets := map[string][]EndpointElem{} // map from subnet/vsi to its nodes/nodeSets from the input
```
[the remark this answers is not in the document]
adisos: comment is still not accurate "map from subnet's/vsi's UID to its nodes/nodeSets from the input"
if it's a vsi, then the map key is ` n.Subnet().UID()` => the uid of the node's subnet 
if it's a subnet, then the map key is `n.VPC().UID() ` => the uid of the subnet's vpc

pkg/vpcmodel/grouping.go
```diff
@@ -219,24 +219,44 @@ func (g *groupingConnections) addPublicConnectivity(ep EndpointElem, commonProps
...
 }
 
-// vsiGroupingBySubnets returns a slice of EndpointElem objects, by grouping set of elements that
-// represent network interface nodes from the same subnet into a single groupedNetworkInterfaces object
-func vsiGroupingBySubnets(groupedConnLines *GroupConnLines, elemsList []EndpointElem) []EndpointElem {
+// vsiOrSubnetsGroupingBySubnetsOrVsis given *GroupConnLines, a list of EndpointElem and a bool saying whether
+// the EndpointElem represents VSIs or subnets.
+// It returns a slice of EndpointElem objects, by grouping the input set of EndpointElem into the input *GroupConnLines
+// such that
+// 1. If the grouped elements are vsis (as by the input bool) then the elements are grouped by their subnet
```
adisos, line 226:
```suggestion
// It returns a slice of EndpointElem objects, by grouping the input set of EndpointElem
// such that
// 1. If the grouped elements are vsis (as by the input bool) then the elements are grouped by their subnet
```

adisos: `groupedConnLines` is used only for caching.. can be changed in a follow up PR , to send only the relevant cache instead of this entire object. (which is misleading)

pkg/vpcmodel/grouping_test.go
adisos: now that we have abstract type for VPC, no need for another mock, we can add the VPC and its uid for the subnets in the test.. I think there is value for covering this functionality in unit tests, and not only rely on end to end tests..

pkg/vpcmodel/grouping.go
```diff
@@ -219,24 +219,44 @@ func (g *groupingConnections) addPublicConnectivity(ep EndpointElem, commonProps
...
+			subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID] = []EndpointElem{}
 		}
-		subnetNameToNodes[subnetName] = append(subnetNameToNodes[subnetName], elem)
+		subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID] = append(subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID], newElem)
 	}
-	for _, nodesList := range subnetNameToNodes {
+	for _, nodesList := range subnetOrVSIToNodesOrNodeSets {
 		if len(nodesList) == 1 { // a single network interface on subnet is just added to the result (no grouping)
 			res = append(res, nodesList[0])
 		} else { // a set of network interfaces from the same subnet is grouped by groupedNetworkInterfaces object
```
adisos, line 262: this comment (by the `else`) also seems outdated

pkg/vpcmodel/grouping.go
ShiriMoran: It will change once we merge with main after we merge the unification PR.

pkg/vpcmodel/grouping.go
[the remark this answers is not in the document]
ShiriMoran: ok, fixed, but without the if.... but with the simpler ()  notation

pkg/vpcmodel/grouping_test.go
ShiriMoran: Did you add any unit test for the abstract `VPC()` and `Subnet()` I can use as reference?

adisos: no, but I think it should be straight forward

ShiriMoran: Not sure what you had in mind, but to my understanding we still need to implement VPCResourceIntf outside ibmvpc; which still requires a mock. I've added mockVPCIntf and restored the test

pkg/vpcmodel/grouping.go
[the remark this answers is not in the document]
adisos: I still don't see how this comment "map from subnet's (vsi's) UID to its nodes (nodeSets) from the input" is accurate... they key is not the subnet/vsi UID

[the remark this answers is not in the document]
adisos: I would rephrase to "a map from vsi/subnet's containing entity UID (subnet/vpc) to the list of EndpointElem objects associated with this key. "

pkg/vpcmodel/grouping.go
ShiriMoran: fixed

pkg/vpcmodel/grouping.go
```diff
@@ -219,57 +219,54 @@ func (g *groupingConnections) addPublicConnectivity(ep EndpointElem, commonProps
...
+		if _, ok := subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID]; !ok {
+			subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID] = []EndpointElem{}
 		}
-		subnetNameToNodes[subnetName] = append(subnetNameToNodes[subnetName], elem)
+		subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID] = append(subnetOrVSIToNodesOrNodeSets[subnetOrVSIUID], newElem)
 	}
-	for _, nodesList := range subnetNameToNodes {
-		if len(nodesList) == 1 { // a single network interface on subnet is just added to the result (no grouping)
+	for _, nodesList := range subnetOrVSIToNodesOrNodeSets {
+		if len(nodesList) == 1 { // a single nif on subnet or vsi on vpc is just added to the result (no grouping)
```
adisos, line 260:
```suggestion
		if len(nodesList) == 1 { // a single nif on subnet or subnet on vpc is just added to the result (no grouping)
```

pkg/vpcmodel/grouping.go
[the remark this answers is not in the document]
ShiriMoran: Right, there is a typo. The key is the Subnet/VPC UID. I'll fix the name and the comment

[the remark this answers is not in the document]
adisos: I see the name has changed, but the comment remained the same..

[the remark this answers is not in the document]
ShiriMoran: fixed

adisos (approved):

PR #425, opened 2024-02-26
Allow using -resource-group and -region flags to filter resources from -vpc-config

olasaadi99:


pkg/ibmvpc/parser.go
```diff
@@ -14,9 +14,96 @@ import (
 	vpcmodel "github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
 )
 
+//nolint:funlen,gocyclo //styles are too long and can not be split
+func FilterByResourceGroup(rc *datamodel.ResourcesContainerModel, resourceGroup string) {
```
zivnevo, line 18:
```suggestion
func filterByResourceGroup(rc *datamodel.ResourcesContainerModel, resourceGroup string) {
```
Should this function be exported?

zivnevo, line 18: I see two problems with this function:
1. It repeats the same code pattern many times
2. If we add a new resource to the collector's data model, we might forget to update this function as well.

I have two possible alternatives:
1. Implement in the collector's data model a function similar to the one here (only solves the second issue)
2. Postpone the filtering a little bit, and implement in `VPCConfigsFromResources` by extending the current `shouldSkipByVPC` mechanism. The nice thing about this alternative is that it can be used to filter by region as well, and that it doesn't duplicate code that much.

Please experiment with these alternatives, and see if they make things any better.

zivnevo (changes requested):

pkg/ibmvpc/parser.go
zivnevo: @adisos , do you have any better idea?

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -538,6 +552,7 @@ func TestUnsupportedAnalysis(t *testing.T) {
 func (tt *vpcGeneralTest) runTest(t *testing.T) {
 	// init test - set the input/output file names according to test name
 	tt.initTest()
+	fmt.Printf("%s\n", tt.resourceGroup)
```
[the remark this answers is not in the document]
olasaadi99: no, it was for debug, my bad

pkg/ibmvpc/parser.go
zivnevo: > I agree this mechanism can be used and extended for this as well, and can save those similar code patterns. But if one runs the analyzer with collection as well (instead of input vpc-config file), and asks to dump the config object into a file, will it be with the filtered resources or not?

Filtering is also applied by the collector in its API calls. This already works.
So, yes - resources will be filtered

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -474,6 +482,13 @@ var tests = []*vpcGeneralTest{
 		grouping:    true,
 		format:      vpcmodel.DRAWIO,
 	},
+	// resource group filtering example
+	{
+		inputConfig:   "multi_resource_groups",
+		useCases:      []vpcmodel.OutputUseCase{vpcmodel.AllEndpoints},
+		format:        vpcmodel.Text,
+		resourceGroup: "ola",
```
adisos, line 490: add in comment what VPCs from the input config file are expected to be filtered out

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -474,6 +482,13 @@ var tests = []*vpcGeneralTest{
 		grouping:    true,
 		format:      vpcmodel.DRAWIO,
 	},
+	// resource group filtering example
```
adisos, line 485: missing also example for filtering by regions

adisos (changes requested):

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -74,6 +76,7 @@ const (
 	secJSONOutSuffix               = "_2nd.json"
 	drawioOutSuffix                = ".drawio"
 	archDrawioOutSuffix            = "_arch.drawio"
+	suffixResourceGroup            = "_resourceGroup_"
```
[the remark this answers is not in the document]
olasaadi99: if no dedicated suffix we will have different files content with the same name to test

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -474,6 +477,22 @@ var tests = []*vpcGeneralTest{
 		grouping:    true,
 		format:      vpcmodel.DRAWIO,
 	},
+	// resource group filtering example
+	// ete-storage-project expected to be filtered out
```
adisos, line 481:
```suggestion
	// ete-storage-project, ete-backup-and-storage expected to be filtered out
```

adisos (commented):
Looks good

adisos (approved):

zivnevo:
Code looks good.
One question I have (also for @adisos ): If we know a VPC has a specific resource group, does it mean everything in the VPC is also from the same resource group? What about resources outside of the VPC (TGWs, LBs)?
Is this handled correctly right now?

adisos:
> Code looks good. One question I have (also for @adisos ): If we know a VPC has a specific resource group, does it mean everything in the VPC is also from the same resource group? What about resources outside of the VPC (TGWs, LBs)? Is this handled correctly right now?

Good question, not sure about whether everything in the VPC is of the same VPC's resource group, we should check that..
TGW is analyzed if it has more than one VPC connected, so if it originally had multiple VPCs but only one is of the requested resource group, then also the TGW will not be analyzed..

LB is not yet supported, so not sure about that.

zivnevo:
> > Code looks good. One question I have (also for @adisos ): If we know a VPC has a specific resource group, does it mean everything in the VPC is also from the same resource group? What about resources outside of the VPC (TGWs, LBs)? Is this handled correctly right now?
> 
> Good question, not sure about whether everything in the VPC is of the same VPC's resource group, we should check that.. TGW is analyzed if it has more than one VPC connected, so if it originally had multiple VPCs but only one is of the requested resource group, then also the TGW will not be analyzed..
> 
> LB is not yet supported, so not sure about that.

But what if the TGW itself has the wrong resource group?

pkg/ibmvpc/parser.go
```diff
@@ -634,12 +634,28 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
...
+	res map[string]*vpcmodel.VPCConfig, resourceGroup string) map[string]*TransitGateway {
 	tgwMap := map[string]*TransitGateway{} // collect all tgw resources
 	for _, tgwConn := range c.TransitConnectionList {
 		tgwUID := *tgwConn.TransitGateway.Crn
 		tgwName := *tgwConn.TransitGateway.Name
 		vpcUID := *tgwConn.NetworkID
+
+		// filtering by resourceGroup
+		if resourceGroup != "" {
+			shouldSkip := true
```
adisos, line 646: This will filter TGW also if the resource group is unknown (the TGW object missing from the config object), and not only if it does not match.
In the test you added "multi_resource_groups" this is the case.. the config object does not contain any TGW objects, only Transit Connection objects..
I would consider at least a warning for filtering due to unknown RG , rather than actual "no match" filtering.

pkg/ibmvpc/examples/out/analysis_out/multi_resource_groups_all_vpcs_.txt
adisos, line 1: when running this test in the cli, there is a gap between the set of warnings in the output when run with `-resource-group` and without it.. please check why

pkg/ibmvpc/parser.go
```diff
@@ -614,12 +634,28 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
...
+			for _, tgw := range c.TransitGatewayList {
+				if tgwUID == *tgw.ID {
+					if *tgw.ResourceGroup.ID == resourceGroup {
+						shouldSkip = false
+					}
```
adisos, line 651:
```suggestion
				if tgwUID == *tgw.ID && *tgw.ResourceGroup.ID == resourceGroup  {
						shouldSkip = false
					}
```

pkg/ibmvpc/parser.go
```diff
@@ -614,12 +634,28 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
...
 	tgwMap := map[string]*TransitGateway{} // collect all tgw resources
 	for _, tgwConn := range c.TransitConnectionList {
 		tgwUID := *tgwConn.TransitGateway.Crn
 		tgwName := *tgwConn.TransitGateway.Name
 		vpcUID := *tgwConn.NetworkID
+
+		// filtering by resourceGroup
+		if resourceGroup != "" {
+			shouldSkip := true
+			for _, tgw := range c.TransitGatewayList {
```
adisos, line 647: instead of iterating `TransitGatewayList` for each `TransitConnection`, you can iterate once before start iterating `TransitGatewayList`, and just keep a map from TGW uid to its RG, to be used later for filtering.

adisos (changes requested):

pkg/ibmvpc/parser.go
olasaadi99: I did split them to break in case tgwUID == *tgw.ID even if *tgw.ResourceGroup.ID != resourceGroup

adisos: oh, ok

pkg/ibmvpc/examples/out/analysis_out/multi_resource_groups_all_vpcs_.txt
olasaadi99: now we have just the unknown RG and just one more warning.. ignoring vpc with no subnets.. the subnets were filtered out

adisos:
```
> ./bin/vpcanalyzer -vpc-config pkg/ibmvpc/examples/input/input_multi_resource_groups.json -resource-group ola
warning: skip fip ete-basic-01 - could not find attached network interface
warning: ignoring floatingIP ete-project-storage-pgw2: target.ResourceType public_gateway is not supported (only networkInterfaceResourceType supported)
warning: ignoring floatingIP agonizing-basket-stubble-deepness: target (FloatingIPTargetIntf) is not of the expected type
warning: ignoring floatingIP zn-vpc1-fip: target (FloatingIPTargetIntf) is not of the expected type
warning: ignoring tgw with unknown RG, tgwID: crn:595
warning: ignoring tgw with unknown RG, tgwID: crn:595
warning: ignoring tgw with unknown RG, tgwID: crn:599
warning: ignoring tgw with unknown RG, tgwID: crn:599
warning: ignoring tgw with unknown RG, tgwID: crn:603
warning: ignoring tgw with unknown RG, tgwID: crn:603
warning: ignoring tgw with unknown RG, tgwID: crn:607
warning: ignoring tgw with unknown RG, tgwID: crn:607

```

1. can avoid multiple warnings on the same ignored tgw  with unknown RG?
2. in the warning msg, change RG to "resource-group" 
3. the warning "warning: skip fip ete-basic-01 - could not find attached network interface" appears only when running with " -resource-group " option .. please check why

pkg/ibmvpc/parser.go
```diff
@@ -636,23 +636,26 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
...
 		// filtering by resourceGroup
 		if resourceGroup != "" {
-			shouldSkip := true
-			for _, tgw := range c.TransitGatewayList {
-				if tgwUID == *tgw.ID {
-					if *tgw.ResourceGroup.ID == resourceGroup {
-						shouldSkip = false
-					}
-					break
+			if RG, ok := tgwIDToRG[tgwUID]; ok {
```
adisos, line 653: rename var `RG` to match Go naming conventions, such as `resourceGroupg`

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -480,6 +483,22 @@ var tests = []*vpcGeneralTest{
 		grouping:    true,
 		format:      vpcmodel.DRAWIO,
 	},
+	// resource group filtering example
+	// // ete-storage-project, ete-backup-and-storage expected to be filtered out
+	{
+		inputConfig:   "multi_resource_groups",
```
adisos, line 489: Missing a test that validates filtering of tgw by resource group, where in the input config file there are multiple TGW objects, some with the input resource-group (that should be analyzed) and some without it (that should be filtered).
This test only covers filtering of tgw due to missing TGW configs.

pkg/ibmvpc/examples/out/analysis_out/multi_resource_groups_all_vpcs_.txt
adisos: for (3):  in a separate PR - consider  storing the information about filtered nodes (uids), so that when parsing fip - can figure out if that is really a required warning, or just a fip configuration that should have been filtered

pkg/ibmvpc/parser.go
```diff
@@ -614,12 +634,32 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
 }
 
 func getTgwObjects(c *datamodel.ResourcesContainerModel,
-	res map[string]*vpcmodel.VPCConfig) map[string]*TransitGateway {
+	res map[string]*vpcmodel.VPCConfig, resourceGroup string) map[string]*TransitGateway {
 	tgwMap := map[string]*TransitGateway{} // collect all tgw resources
+
+	tgwIDToRG := map[string]string{}
```
zivnevo, line 640: How about having a `map[string]bool` that keeps a set of TGWs to skip?

pkg/ibmvpc/parser.go
```diff
@@ -614,12 +634,32 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
 }
 
 func getTgwObjects(c *datamodel.ResourcesContainerModel,
-	res map[string]*vpcmodel.VPCConfig) map[string]*TransitGateway {
+	res map[string]*vpcmodel.VPCConfig, resourceGroup string) map[string]*TransitGateway {
```
zivnevo, line 637: Should we also consider the region (`Location`) of the TGW?

pkg/ibmvpc/analysis_output_test.go
olasaadi99: updated the test

pkg/ibmvpc/parser.go
olasaadi99: @adisos ?

adisos: yes, should probably filter by Location of TGW if the input region was given.

pkg/ibmvpc/parser.go
```diff
@@ -613,13 +633,51 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
...
+			if tgw, ok := tgwIDToTgw[tgwUID]; ok {
+				if tgw == nil || !slices.Contains(regions, *tgw.Location) {
+					continue
+				}
+			} else {
+				fmt.Printf("warning: ignoring tgw with unknown region, tgwID: %s\n", tgwUID)
+				tgwIDToTgw[tgwUID] = nil // to avoid having this tgw's same warning issued again from another transitConnection
+				continue
+			}
+		}
```
adisos, line 680: 1. As Ziv suggested, better have a `map[string]bool` that keeps a set of TGWs to skip.. + avoid current code duplication for warnings handling 
2. missing tests for filtering of TGW by region as well

adisos (changes requested):

pkg/ibmvpc/analysis_output_test.go
adisos: great, looks good. 
please also update the comment in the test file about the expected TGW objects filtered out by resource group.

pkg/ibmvpc/parser.go
olasaadi99: need this map anyway.. added the other map

pkg/ibmvpc/examples/out/analysis_out/multi_resource_groups_all_vpcs_.txt
adisos: please open a separate issue for the comment above

pkg/ibmvpc/parser.go
```diff
@@ -613,13 +633,58 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
...
+	if resourceGroup != "" || len(regions) > 0 {
+		tgwIDToTgw = getTgwMap(c)
+	}
+
 	for _, tgwConn := range c.TransitConnectionList {
 		tgwUID := *tgwConn.TransitGateway.Crn
 		tgwName := *tgwConn.TransitGateway.Name
 		vpcUID := *tgwConn.NetworkID
+
+		if toSkip, ok := tgwToSkip[tgwUID]; ok && toSkip {
```
adisos, line 658: why `&& toSkip` ? any key added to this map has value `true`

pkg/ibmvpc/parser.go
```diff
@@ -613,13 +633,58 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
 	return nil
 }
 
+func getTgwMap(c *datamodel.ResourcesContainerModel) map[string]*datamodel.TransitGateway {
+	tgwIDToTgw := map[string]*datamodel.TransitGateway{}
+	for _, tgw := range c.TransitGatewayList {
```
adisos, line 638: if already iterating `c.TransitGatewayList`, can compute `tgwToSkip` here?

olasaadi99: no, this map is to skip and not warn about the skipped transit gateway connection again

pkg/ibmvpc/examples/out/analysis_out/multi_resource_groups_all_vpcs_.txt
olasaadi99: opened https://github.com/np-guard/vpc-network-config-analyzer/issues/440

adisos (approved):
LGTM

PR #433, opened 2024-03-06
387 explain by path

ShiriMoran:


ShiriMoran:
I did some reorg of the printing code, renaming functions, adding documentation add moving stuff around.
However, apart of the new path related code no functional code was changed

PR #438, opened 2024-03-07
if the input is not a legal cidr/IP address then the error message sh…

ShiriMoran:
…ould be broader than the error just not being a legal cidr/IP address


adisos (approved):
looks good

ShiriMoran:
@zivnevo personally I found the bug fixed by this PR very confusing. If you want to consider a release between V3.0 to V4.0

PR #441, opened 2024-03-10
368 explainability iks node

ShiriMoran:


pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -33,12 +33,14 @@ type srcDstDetails struct {
 	// filters relevant for this src, dst pair; map keys are the filters kind (NaclLayer/SecurityGroupLayer)
 	// for two internal nodes within same subnet, only SG layer is relevant
 	// for external connectivity (src/dst is external) with FIP, only SG layer is relevant
-	filtersRelevant     map[string]bool
-	potentialAllowRules *rulesConnection // potentially enabling connection - potential given the filter is relevant
-	actualAllowRules    *rulesConnection // enabling rules effecting connection given router; e.g. NACL is not relevant for fip
-	potentialDenyRules  *rulesConnection // deny rules potentially (w.r.t. router) effecting the connection, relevant for ACL
-	actualDenyRules     *rulesConnection // deny rules effecting the connection, relevant for ACL
-	actualMergedRules   *rulesConnection // rules actually effecting the connection (both allow and deny)
+	// for iks node sg is not relevant - this is the reason we need separation between ingress and egress
```
adisos, line 36: I don't think it's a correct statement that SG is not relevant for IKS nodes, only we cannot obtain it currently in the input config object. 
So I wouldn't change the code and modeling with such assumption.
Instead, maybe we can for now add a default (dummy) SG that allows all?
So that when we do have the SG support, it is a smaller update? (and the code is more accurate in modeling aspects).

ShiriMoran: How do you want to add this default SG? and to where exactly?

Adding an SG that does not exist and reporting its rules as enabling could be highly confusing. 
Why do we report iks-nodes related results when we can't read the relevant SG and don't actually know what are the results?

adisos: > How do you want to add this default SG? and to where exactly?

maybe in the parser, when building the vpc config object

> Adding a SG that does not really exist and reporting its rules as enabling 

I don't suggest to report that SG or describe it as enabling, for explainability we can ignore this SG in the report, and add a message that analysis lacks SG information 


> Why do we report iks-nodes related results when we can't read the relevant SG

We need to analyze IKS configs

adisos: @zivnevo @kyorav  what do you think?

kyorav: IKS automatically apply two SGs to all worker nodes:
1. the VPC's default SG (you can find it in the config object, under the VPC config)
2. a security group that they generate, which is called "kube-<cluster-ID>"
I think we should do the analysis with both of those.

It is possible for the user to request additional SGs at creation time, which we wouldn't know about, which makes our analysis an under approximation. We can add a warning message that mentions that.

ShiriMoran: If the explainability is going to have a dedicated mechanism for this case - as the one in this PR (which we can of course discuss), then what's the point of adding a non-existing SG which will be ignored anyways?

adisos: right, no need to add a non-existing SG
I think that in the parser, when adding IKS nodes from the config objects, should also handle that and associate the relevant SGs from the config object to the IKS nodes.

adisos: opened issue #443

PR #451, opened 2024-03-12
add region field to the resources model

olasaadi99:


adisos (changes requested):
few comments:
- I think it should be sufficient to keep the region per VPC resource, and per TGW.. note that any resource implements the VPCResourceIntf, that has a reference to the VPC it belongs to..
so no need to propagate in parser the Region through the different resource. 
- I would expect that `VPCResourceIntf` would have a method `Region()` 
- why are input test files changed?

haim-kermany (commented):
Hi
for later drawio implementation, we need that:
1. the region will be represent as a struct.
2. A region should only one instance of this struct

olasaadi99:
> * why are input test files changed?
in json files output we print resources details, that includes the additional field region, so it printed region: "", so I just added a region field in the input to print it

adisos:
Summary from discussion:
- add new type `Region` in pkg `ibmvpc`, and methods `Region() *Region` for types `VPC` and `TransitGateway` within `ibmvpc`
- For abstract vpc can have a method `RegionName() string` (similar to `ZoneName() string` )

pkg/ibmvpc/vpc.go
```diff
@@ -21,6 +21,24 @@ func getNodeName(name, addr string) string {
 	return fmt.Sprintf("%s[%s]", name, addr)
 }
 
+type Region struct {
+	name string
+}
+
+var (
+	regionToStructMap = make(map[string]*Region)
```
adisos, line 25: Why is this cache defined globally? I think it can defined in the function `VPCConfigsFromResources`

pkg/ibmvpc/parser.go
```diff
@@ -744,6 +730,7 @@ func getTgwObjects(c *datamodel.ResourcesContainerModel,
 				},
 				vpcs:            []*VPC{vpc},
 				availableRoutes: map[string][]*ipblocks.IPBlock{},
+				region:          &Region{name: region},
```
adisos, line 733: should use `regionToStruct` here ?
also, use `*tgw.Location` (var `region` is not always assigned)

adisos (changes requested):

pkg/ibmvpc/parser.go
```diff
@@ -675,10 +690,11 @@ func getTgwObjects(c *datamodel.ResourcesContainerModel,
 		if _, ok := tgwToSkip[tgwUID]; ok {
 			continue
 		}
+		tgw, ok := tgwIDToTgw[tgwUID]
```
adisos, line 693:
```suggestion
		tgw, hasTgwConfig:= tgwIDToTgw[tgwUID]
```
Better use meaningful name here .
Another option is to have just `tgw = tgwIDToTgw[tgwUID]` , and check if `tgw` is `nil` to check if it has no config

pkg/ibmvpc/parser.go
```diff
@@ -708,14 +725,20 @@ func getTgwObjects(c *datamodel.ResourcesContainerModel,
 			continue
 		}
 		if _, ok := tgwMap[tgwUID]; !ok {
+			region := ""
+			if ok { // if there is a transit gateway in the config file
```
adisos, line 729: this `ok` (line `729`) refers the one defined from line `727` instead of the one you meant (from line `693`)

pkg/ibmvpc/parser.go
adisos: also, better rename `tgw` to `tgwFromConfig`  (line 732 defines another var named `tgw`, and this may be confusing)

adisos (changes requested):

adisos (commented):
consider adding a test, that also checks the returned `*Region` object from the methods added to `Vpc` and `TransitGateway`

pkg/ibmvpc/vpc.go
```diff
@@ -17,6 +17,20 @@ func getNodeName(name, addr string) string {
 	return fmt.Sprintf("%s[%s]", name, addr)
 }
 
+type Region struct {
+	name string
+}
+
+func regionToStruct(regionName string, regionToStructMap map[string]*Region) *Region {
```
haim-kermany, line 24: 1. rename it to getRegionByName()
2. I think it should be at parser.go. @adisos?

adisos: yes, I also think so

pkg/ibmvpc/parser.go
```diff
@@ -30,6 +30,16 @@ func ParseResourcesFromFile(fileName string) (*datamodel.ResourcesContainerModel
 	return &config, nil
 }
 
+func getRegionByName(regionName string, regionToStructMap map[string]*Region) *Region {
```
adisos, line 33: if `regionName` is empty string, return `nil`? @haim-kermany  please confirm

haim-kermany: Please return a Region with empty name, not nil.
Or there should be a default name, @adisos ?

olasaadi99: that is what we do right now.. if regionName is empty string we return region struct with empty name ""
no there is no default name

pkg/ibmvpc/parser_test.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+func TestVPCResourceModelRegion(t *testing.T) {
+	rc, err := ParseResourcesFromFile(filepath.Join(getTestsDirInput(), "input_experiments_env.json"))
+	require.Nilf(t, err, "err: %s", err)
+	vpcConfigs, err := VPCConfigsFromResources(rc, "", "", nil, false)
+	require.Nilf(t, err, "err: %s", err)
+	vpcConfig := vpcConfigs["crn:17"]
+	require.Equal(t, vpcConfig.VPC.RegionName(), "us-south")
+}
```
adisos, line 20: I would extend this function, to also have input from config json file with transit gateways, 
and check that the generated TransitGateway object in the returned VPCConfig, returns an expected Region object when calling `tgw.Region()`.  Similarly, can check that the VPC object it holds, is as expected with `vpc.Region()`.

pkg/ibmvpc/parser_test.go
```diff
@@ -0,0 +1,68 @@
...
+	regionToStructMap := make(map[string]*Region) // map for caching Region objects
+	getRegionByName("us-south", regionToStructMap)
+	getRegionByName("us-south", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+
+	require.Equal(t, len(regionToStructMap), 2)
+	require.Equal(t, getRegionByName("us-south", regionToStructMap), getRegionByName("us-south", regionToStructMap))
```
adisos, line 67: from the documentation of function `Equal`: "Pointer variable equality is determined based on the equality of the referenced values (as opposed to the memory addresses)."

so would consider also testing that same memory address is returned.

pkg/ibmvpc/parser_test.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func TestGetRegionByName(t *testing.T) {
+	regionToStructMap := make(map[string]*Region) // map for caching Region objects
+	getRegionByName("us-south", regionToStructMap)
+	getRegionByName("us-south", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
+	getRegionByName("us-east", regionToStructMap)
```
adisos, line 64: better also add a test with an empty region name

adisos (changes requested):
looks good, just few more comments

pkg/ibmvpc/parser_test.go
```diff
@@ -0,0 +1,80 @@
...
+	getRegionByName("", regionToStructMap)
+	require.Equal(t, len(regionToStructMap), 3)
+
+	region1 := getRegionByName("us-south", regionToStructMap)
+	region2 := getRegionByName("us-south", regionToStructMap)
+	require.Equal(t, region1, region2)
+	require.Equal(t, region1 == region2, true)
```
adisos, line 79:
```suggestion
	require.True(t, region1 == region2)
```

adisos (commented):
looks good, just one small comment

haim-kermany (approved):

adisos (approved):

PR #463, opened 2024-03-18
Bump models to v0.2.1

elazarg:
Some changes are related to loop variable aliasing that I think are unrelated to the bump. 

This removes the dependency on vpf-network-config-synthesis.


pkg/ibmvpc/parser.go
```diff
@@ -562,7 +563,8 @@ func getSGconfig(rc *datamodel.ResourcesContainerModel,
 ) error {
 	sgMap := map[string]map[string]*SecurityGroup{} // map from vpc uid to map from sg name to its sg object
 	sgLists := map[string][]*SecurityGroup{}
-	for _, sg := range rc.SecurityGroupList {
+	for i := range rc.SecurityGroupList {
+		sg := rc.SecurityGroupList[i]
```
zivnevo, line 567: Why is this better?

pkg/ibmvpc/parser.go
```diff
@@ -622,7 +624,8 @@ func getNACLconfig(rc *datamodel.ResourcesContainerModel,
 	skipByVPC map[string]bool,
 ) error {
 	naclLists := map[string][]*NACL{} // map from vpc uid to its nacls
-	for _, nacl := range rc.NetworkACLList {
+	for i := range rc.NetworkACLList {
+		nacl := rc.NetworkACLList[i]
```
zivnevo, line 628: same here

zivnevo (approved):
Just these two tiny comments. Feel free to ignore

pkg/ibmvpc/parser.go
elazarg: Same warning:
```golang
naclAnalyzer, err := NewNACLAnalyzer(&nacl.NetworkACL)
```

pkg/ibmvpc/parser.go
zivnevo: It's a linter false-positive. If I add parenthesis like:
```golang
			analyzer: NewSGAnalyzer(&(sg.SecurityGroup)), members: map[string]vpcmodel.Node{},
```
the linter is ok with it.

elazarg: Are you 100% sure it's a false positive? The parenthesis could be interfering with the syntactic rules of the linter.

The right solution is probably moving to golang 1.22 in which implicit memory aliasing issue is gone. For now, I vote for keeping it an explicit variable.

zivnevo: Well, `rc.SecurityGroupList` is a slice of pointers, so `sg` is assigned to point to `rc.SecurityGroupList[i]`. I don't see how this is different from your version of the code.
The linter warning makes sense when iterating over slices of **structs**. In this case, the variable is only a temporary copy of the struct in the slice, and taking its address is almost always an error.

elazarg: What do you want to do with this?

zivnevo: Just merge it with your version of the code

PR #464, opened 2024-03-18
454 explain multivpc input with vpc

ShiriMoran:


haim-kermany (approved):

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -224,21 +212,34 @@ func (c *VPCConfig) getNodesFromInputString(cidrOrName string) (nodes []Node, in
...
 
 // getNodesOfVsi gets a string name or UID of VSI, and
 // returns the list of all nodes within this vsi
-func (c *VPCConfig) getNodesOfVsi(vpc, vsi string) ([]Node, int, error) {
+func (c *VPCConfig) getNodesOfVsi(name string) ([]Node, int, error) {
 	var nodeSetWithVsi NodeSet
+	// vsi name may be prefixed by vpc name
+	var vpc, vsi string
+	uid := name // uid specified - vpc prefix is not relevant and uid may contain "/"
+	cidrOrNameSlice := strings.Split(name, "/")
```
adisos, line 225: add const for the "/" separator

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -224,21 +212,34 @@ func (c *VPCConfig) getNodesFromInputString(cidrOrName string) (nodes []Node, in
...
+	switch len(cidrOrNameSlice) {
+	case 1: // vpc name not specified
+		vsi = name
+	case 2: // vpc name specified
+		vpc = cidrOrNameSlice[0]
+		vsi = cidrOrNameSlice[1]
+	default: // > 1, address
+		vsi = name
+	}
```
adisos, line 234: why do you differentiate case `1` and `default` here? 
where is `address` assumed?

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -47,7 +47,7 @@ const noValidInputMsg = "does not represent a legal IP address, a legal CIDR or
 // getVPCConfigAndSrcDstNodes given src, dst names returns the config in which the exaplainability analysis of these
 // should be done and the Nodes for src and dst.
 
-// src and dst when referring to a vsi *name* may be prefixes with their vpc name to solve ambiguity
+// src and dst when referring to a vsi *name* may be prefixes "/" with their vpc name to solve ambiguity
```
adisos, line 50: documentation should also be added to the cli args, for valid input format of vsi of the explainability option

pkg/vpcmodel/explainabilityInput.go
ShiriMoran: the address is from the time I used "."
No longer relevant, removed.

pkg/vpcmodel/explainabilityInput.go
ShiriMoran: right. done.

cmd/analyzer/parse_args.go
```diff
@@ -151,7 +151,8 @@ var supportedAnalysisTypesList = []string{
 	explainMode,
 }
 
-const srcDstUsage = "endpoint for explanation; can be specified as a VSI name/CRN or an internal/external IP-address/CIDR"
+const srcDstUsage = "endpoint for explanation; can be specified as a VSI name/CRN or an internal/external IP-address/CIDR" +
+	"in multivpc context ambiguity in VSI names can be resolves by prefixing the VPC name with /"
```
adisos, line 155:
```suggestion
	"VSI name can be specified as <vsi-name> or  <vpc-name>/<vsi-name>"
```

pkg/vpcmodel/explainabilityInput.go
adisos: shouldn't issue an error in this case?

ShiriMoran: no. uid may contain more than one such deliminator since its a `crn`. btw, I still don't see why `UID` is `crn` and not `id`.

Issue #387, opened 2024-02-14
Explainability output enhancements 


adisos:
Example for current output: 
```
Connectivity explanation for protocol: TCP between vsi1-ky[10.240.10.4] and 161.26.0.0/16
===================================================================================================

Connection protocol: TCP exists between vsi1-ky[10.240.10.4] and Public Internet 161.26.0.0/16
External Router PublicGateway: public-gw-ky
Egress: security group sg1-ky allows connection (with allow rules); network ACL acl1-ky allows connection (with allow rules)
```

I thought we could improve the output by presenting the actual "path" of the relevant resources/configurations affecting the connectivity, by the relevant order. 

so for the case above, it may be like:
```
vsi1-ky[10.240.10.4] ->  SG sg1-ky ->  ACL acl1-ky -> PublicGateway: public-gw-ky ->  Public Internet 161.26.0.0/16
```

or also with the subnet:
```
vsi1-ky[10.240.10.4] ->  SG sg1-ky -> subnet ... ->  ACL acl1-ky -> PublicGateway: public-gw-ky ->  Public Internet 161.26.0.0/16
```

For denied connectivity, could be a prefix of the path until the part that blocks the connectivity. 

@ShiriMoran @kyorav @zivnevo  what do you think?

kyorav:
I think it's a very good idea. It makes it much easier to understand the explanation and removes a lot of unnecessary verbiage. 
In terms of format, there are a few corner cases we need to figure out. For example, if there are multiple SGs allowing the connection on the same endpoint we need to list them all in the same "step". So maybe put them in brackets?
```
vsi1-ky[10.240.10.4] ->  SG [sg1-ky, sg2-ky] -> subnet ...
```
Also, this puts everything in one long line, which can be hard to follow and will look differently in different displays (with/without line wrap of different size). Maybe we can put each step in a separate line, and then we can also use indentation to convey some information
```
vsi1-ky[10.240.10.4] ->  
    SG sg1-ky ->  
    ACL acl1-ky -> 
    PublicGateway: public-gw-ky ->  
Public Internet 161.26.0.0/16
```

zivnevo:
I think this will become really useful as we add more and more network devices (RTs, LBs) and as we move to multi-cloud scenarios.
However, I don't want to delay v0.3 for this feature. Maybe make it for v0.4 or even v0.5.

ShiriMoran:
[...]

PR #472, opened 2024-03-20
387 add external ip to fip

ShiriMoran:


ShiriMoran:
solves last request in #387

zivnevo (approved):
LGTM

PR #482, opened 2024-03-26
fix in external node cach 

haim-kermany:
need to give pointers in the cash


ShiriMoran:
Looks good. Can you add a test derived from the example in which it did not work before the fix? @haim-kermany

haim-kermany:
> Looks good. Can you add a test derived from the example in which it did not work before the fix? @haim-kermany

the drawio test already on it

ShiriMoran (approved):

PR #485, opened 2024-03-28
associate the relevant SGs to the IKS nodes

olasaadi99:


adisos:
https://github.com/np-guard/vpc-network-config-analyzer/blob/b51a1dee6e3c200b9349d89bda5973a3d7cfbb92/pkg/ibmvpc/vpc.go#L501-L503 

Should remove the lines from the reference above, so that connectivity analysis does not assume anymore that no SG is associated with these nodes.

pkg/ibmvpc/parser.go
```diff
@@ -157,6 +157,7 @@ const (
 	networkInterfaceResourceType = "network_interface" // used as the type within api objects (e.g. SecurityGroup.Targets.ResourceType)
 	vpeResourceType              = "endpoint_gateway"  // used as the type within api objects (e.g. SecurityGroup.Targets.ResourceType)
 	loadBalancerResourceType     = "load_balancer"     // used as the type within api objects (e.g. SecurityGroup.Targets.ResourceType)
+	iksNodeResourceType          = "iks_node"          // used as the type within api objects (e.g. SecurityGroup.Targets.ResourceType)
```
adisos, line 160:
```suggestion
       // iksNodeResourceType is not actually used from input api objects, but is added by the parser to SGs with targets
       // that should be added with iks nodes  
	iksNodeResourceType          = "iks_node"          // used as the type within api objects (e.g. SecurityGroup.Targets.ResourceType)
```

pkg/ibmvpc/parser.go
```diff
@@ -995,11 +1002,51 @@ func getSubnetByCidr(res vpcmodel.MultipleVPCConfigs, cidr string) (*Subnet, err
...
+	}
+	return nil
+}
+
+func findDefaultSGForVpc(rc *datamodel.ResourcesContainerModel, vpcUID string) *datamodel.SecurityGroup {
+	for _, vpc := range rc.VpcList {
+		if *vpc.CRN != vpcUID {
+			continue
+		}
+		sgCRN := vpc.DefaultSecurityGroup.CRN
```
adisos, line 1019: rename `sgCRN` to `DefaultSgCRN`

pkg/ibmvpc/parser.go
```diff
@@ -1029,6 +1080,9 @@ func getIKSnodesConfig(res vpcmodel.MultipleVPCConfigs,
 					SubnetResource: subnet,
 				},
 			}
+			res[vpcUID].UIDToResource[nodeObject.ResourceUID] = nodeObject
+			addIKSNodeAsSGTarget(sg, iksCluster)
```
adisos, line 1084: why is `addIKSNodeAsSGTarget` called foe each worker node, if it iterates all worker nodes in the cluster and adds them as SG targets? 
Shouldn't it be called outside the loop?

olasaadi99: because we associate each iks worker node to it's SG not each iks cluster

adisos: so why on each worker node you call `addIKSNodeAsSGTarget` with `iksCluster`, and then within this function you iterate all cluster nodes?

pkg/ibmvpc/parser.go
adisos: please change to `defaultSgCRN`, my bad..

pkg/ibmvpc/parser.go
olasaadi99: right should be after the sec loop

pkg/ibmvpc/parser.go
```diff
@@ -1015,6 +1064,10 @@ func getIKSnodesConfig(res vpcmodel.MultipleVPCConfigs,
 				continue
 			}
 			vpcUID := subnet.VPC().UID()
+			if i == 0 {
+				// first iksNode
```
adisos, line 1068:
```suggestion
				// first iksNode - assuming all cluster nodes are in the same vpc, thus sufficient to check vpc of the first node 
```

pkg/ibmvpc/parser.go
```diff
@@ -995,11 +1004,51 @@ func getSubnetByCidr(res vpcmodel.MultipleVPCConfigs, cidr string) (*Subnet, err
...
+		resourceType := iksNodeResourceType
+		target := &vpc1.SecurityGroupTargetReference{
+			ID:           iksNode.ID,
+			ResourceType: &resourceType,
+		}
+		sg.Targets = append(sg.Targets, target)
+	}
+}
+
 func getIKSnodesConfig(res vpcmodel.MultipleVPCConfigs,
```
adisos, line 1045:
```suggestion
// assuming getIKSnodesConfig is called before getSGconfig, becuase it updates the input SG targets with missing IKS nodes, if there are such 
func getIKSnodesConfig(res vpcmodel.MultipleVPCConfigs,
```

adisos (commented):
Just few more documentation to add

adisos (approved):
LGTM

PR #487, opened 2024-03-31
Drawio routers w grouping

haim-kermany:
handle the case that src/dst is a groupedEndpointsElems.
in this case, we need to take the router from the group elements

### DIFF
--- a/pkg/vpcmodel/drawioOutput.go
+++ b/pkg/vpcmodel/drawioOutput.go
@@ -151,7 +151,6 @@ func (d *DrawioOutputFormatter) lineRouter(line *groupedConnLine, vpcResourceNam
 			}
 		}
 		return d.nodeRouters[d.gen.TreeNode((*group)[0])]
-
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/vpcmodel/drawioOutput.go
```diff
@@ -132,13 +132,26 @@ func (d *DrawioOutputFormatter) createRouters() {
...
 	case line.src.IsExternal():
-		return d.nodeRouters[d.gen.TreeNode(line.dst)]
+		routeredEP = line.dst
+	default:
+		return nil
+	}
+	if group, ok := routeredEP.(*groupedEndpointsElems); ok {
+		firstRouter := d.nodeRouters[d.gen.TreeNode((*group)[0])]
+		for _, node := range *group {
+			if d.nodeRouters[d.gen.TreeNode(node)] != firstRouter {
```
ShiriMoran, line 150: Note that we can add, optionally with a bool flag,  the router to the grouping key, in which case we will not have to return nil in this case. Returning nil when a Router exists can be misleading, I guess.

ShiriMoran (approved):
LGTM 
see comment

pkg/vpcmodel/drawioOutput.go
haim-kermany: not sure I understand

adisos (approved):

PR #494, opened 2024-04-03
update makefile

adisos:


Makefile
```diff
@@ -23,4 +23,7 @@ build:
 
 test:
 	@echo -- $@ --
-	go test ./... -v -cover -coverprofile analyzer.coverprofile
+	 go test ./... -v -coverpkg=./... -coverprofile analyzer.coverprofile
```
zivnevo, line 26: Maybe define a variable to contain `analyzer.coverprofile` and use it in the two directives

adisos: done

zivnevo (approved):

Issue #412, opened 2024-02-21
Grouping should not be based on names - switch to UID()


ShiriMoran:
Theoretically basing grouping on names was always a bad idea, as names are not unique.
Practically, it is especially a bad idea in the context of multi vpcs.

adisos:
>  as names are not unique.

The names of network interfaces in the model consists of vsi name + ip address, so it is less likely to have multiple identical names across different VPCs. 

Still, better change to `UID`.

ShiriMoran:
> > as names are not unique.
> 
> The names of network interfaces in the model consists of vsi name + ip address, so it is less likely to have multiple identical names across different VPCs.
> 
> Still, better change to `UID`.

1. Once we support multiple vpc with same addresses prefix space it will be more likely
2. In subnet analysis mode the keys are subnets names. Not sure how likely it is that subnets across VPC will share names (probably it is not likely for subnets of the same VPC to share names). But it is possible

Having said that, I'm fine with giving this low priority @zivnevo @adisos

ShiriMoran:
solved by https://github.com/np-guard/vpc-network-config-analyzer/pull/461

PR #495, opened 2024-04-03
412 grouping uid

ShiriMoran:
solves #412


pkg/ibmvpc/analysis_output_test.go
```diff
@@ -515,6 +515,13 @@ var tests = []*vpcGeneralTest{
 		grouping:    true,
 		format:      vpcmodel.Text,
 	},
+	// grouping test of identical names different names that should not be grouped
```
adisos, line 518: what do you mean by `identical names different names` ?

pkg/ibmvpc/examples/out/analysis_out/tgw_basic_example_dup_subnet_names_all_vpcs_subnetsBased_withPGW_with_grouping.txt
```diff
@@ -0,0 +1,9 @@
+Connectivity between VPCs connected by TGW local-tg-ky (UID: crn:161)
+ky-vpc1-net1 => ky-vpc1-net1 : All Connections
+ky-vpc1-net1 => ky-vpc1-net1 : All Connections
```
adisos, line 3: open an issue with reference to this example output, where subnet name should also include vpc , at least in the cross-vpc connectivity ?

ShiriMoran: Sure.  https://github.com/np-guard/vpc-network-config-analyzer/issues/498
But note that we can also have two subnets with the same name in the same VPC (though one may claim in is less likely); As I have used this input file to demonstrate the dup names issue I could have chosen another one with only one VPC.
If we really want to tackle dup names we should perhaps issue some lint error. Or have some renaming mechanism? perhaps open issue for this as well?

adisos: ok, updated the issue to contain these as two related tasks

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -515,6 +515,13 @@ var tests = []*vpcGeneralTest{
 		grouping:    true,
 		format:      vpcmodel.Text,
 	},
+	// grouping test of identical names different resources and thus different UIDs that should not be grouped
+	{
+		inputConfig: "tgw_basic_example_dup_subnet_names",
+		useCases:    []vpcmodel.OutputUseCase{vpcmodel.AllSubnets},
+		grouping:    true,
+		format:      vpcmodel.Text,
+	},
```
adisos, line 524: Is this test supposed to fail before the change in the grouping code? Isn't the grouping prevented for cross-vpc subnets anyway?

ShiriMoran: No, grouping is not prevented in cross-vpc subnets:
E.g.:
```
Connectivity between VPCs connected by TGW local-tg-ky (UID: crn:578)
subnet11-ky,subnet12-ky => subnet21-ky : All Connections
subnet21-ky => subnet11-ky,subnet12-ky : All Connections
subnet21-ky => subnet31-ky,subnet32-ky : All Connections
subnet31-ky => subnet11-ky,subnet12-ky : All Connections *
subnet31-ky,subnet32-ky => subnet21-ky : All Connections
```

ShiriMoran: Done, I also verified that the test fails in main.

Btw, this issue is relevant only to subnets with the same name within the same vpc. Perhaps we should have given it lower priority. Anyways now its fixed.

adisos (approved):

PR #497, opened 2024-04-03
Parser updates

adisos:
Few updates as preparation for #496 : 
- parser updates: separate functions that return the created resources, so that can be reused in unit-tests. 
- vpc struct update: add field `addressPrefixesIPBlock`

### DIFF
--- a/pkg/ibmvpc/parser.go
+++ b/pkg/ibmvpc/parser.go
@@ -211,6 +211,28 @@ func updateFilteredOutNetworkInterfacesUIDs(instance *datamodel.Instance, filter
 	}
 }
 
+func newNetworkInterface(name, uid, zone, address, vsi string, vpc vpcmodel.VPCResourceIntf) (*NetworkInterface, error) {
+	intfNode := &NetworkInterface{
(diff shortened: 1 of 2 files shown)


pkg/ibmvpc/parser.go
```diff
@@ -506,6 +531,25 @@ func getVPCAddressPrefixes(vpc *datamodel.VPC) (res []string) {
...
+			ResourceType: ResourceTypeVPC,
+			Region:       region,
+		},
+		nodes:           []vpcmodel.Node{},
+		zones:           map[string]*Zone{},
+		addressPrefixes: ap,
+		region:          getRegionByName(region, regionToStructMap),
+	}
+
+	vpcNodeSet.addressPrefixesIPBlock, _ = ipblock.FromCidrList(ap)
```
zivnevo, line 548: We don't ignore errors in building IPBlocks in the other ctors. Why here?

adisos: right, forgot this one.. added error handling there

zivnevo (approved):

PR #501, opened 2024-04-07
489 explain no tgw

ShiriMoran:


ShiriMoran:
replaced by https://github.com/np-guard/vpc-network-config-analyzer/pull/502

Issue #489, opened 2024-04-01
bug in explainability multi-vpc when referring to src/dst with vpc name prefix


ShiriMoran:
Currently src, dst that are defined within the same vpc (or multivpc) is looked for as pre-condition.
But if a tgw between the src and dst is not defined then the result is that src, dst not found and an input error is asserted.
Fix:
In case src, dst each found in a different VPC (but not in the same one) identify the case as no tgw defined.

After the fix, can check the following:
The following cmd line refers to existing src, dst yet get an error message:
`... -analysis-type explain -format debug -src test-vpc0-ky/vsi1-ky -dst test-vpc1-ky/vsi1-ky`
error message:
2024/04/01 12:35:55 illegal src: test-vpc0-ky/vsi1-ky does not represent a legal IP address, a legal CIDR or a VSI name. exiting...
exit status 1

Also, check why the following does not find src in the same input:
`-analysis-type explain -format debug -src 10.240.3.5 -dst 10.240.12.4`

ShiriMoran:
solved by #502

PR #502, opened 2024-04-07
489 explain no tgw

ShiriMoran:


ShiriMoran:
solves #489

ShiriMoran:
replaces by #512

Issue #467, opened 2024-03-19
can now check address prefixes of the vpc instead of subnets


ShiriMoran:
can now check address prefixes of the vpc instead of subnets

_Originally posted by @adisos in https://github.com/np-guard/vpc-network-config-analyzer/pull/458#discussion_r1528677395_

ShiriMoran:
I'm not sure this has to be done -  I think the current error messages that refers to subnets is good enough

ShiriMoran:
@zivnevo I think this should not be part of V0.4

zivnevo:
I cannot understand what this issue is about from its title + description, so I cannot decide if it should be part of 0.4 or not.
@adisos , if you understand the context, please set the milestone.

ShiriMoran:
This is about the error messages delivered in explainability. The question was whether to check, for a given vsi address as src/dst, if its inside the subnet's address range or the VPC's address prefix.
Anyways, I've just noticed this was already solved - I'll just update the relevant error messages and submit a PR. So we can keep in in V0.4

ShiriMoran:
Turns out there are config objects that are missing the input address prefixes. ROI of having a WA for this not worth it. We'll stay with the subnets range for now.

PR #503, opened 2024-04-07
467 ap errormsg

ShiriMoran:


ShiriMoran:
solves #467

ShiriMoran:
contained in #502

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -327,12 +327,12 @@ func (c *VPCConfig) getNodesFromAddress(ipOrCidr string, inputIPBlock *ipblock.I
 		if !inputIPBlock.ContainedIn(vpcAP) {
```
adisos, line 327: `vpcAP` is computed from `c.VPC.AddressRange()` instead of `c.VPC.AddressPrefixes()`

```
func (v *VPC) AddressRange() *ipblock.IPBlock {
	return v.internalAddressRange
}
```

// internalAddressRange is the union of all the vpc's subnets' CIDRs

adisos (changes requested):

pkg/vpcmodel/explainabilityInput.go
ShiriMoran: Right. Why does `c.VPC.AddressPrefixes()` returns a string?
Can an address prefix be more than one `CIDR`? I believe it can.

adisos: It returns a slice of strings:
```
func (v *VPC) AddressPrefixes() []string {
	return v.addressPrefixes
}
```

you can use ipblock package to get IPBlock from list of CIDR strings

adisos: this is for a separate issue, to consider this change

ShiriMoran: #505

ShiriMoran: Turns out there are config objects that are missing the input address prefixes. ROI of having a WA for this not worth it. We'll stay with the subnets range

PR #510, opened 2024-04-09
Add logger for info/warning/error messages handling

olasaadi99:


README.md
```diff
@@ -10,8 +10,8 @@ $ ./bin/vpcanalyzer -h
 Usage of vpc-network-config-analyzer:
   -analysis-type string
         Supported analysis types:
-        * all_endpoints  - supported with: txt, md, json, drawio, arch_drawio, debug
-        * all_subnets  - supported with: txt, md, json, drawio, arch_drawio
+        * all_endpoints  - supported with: txt, md, json, drawio, arch_drawio, svg, arch_svg, html, arch_html, debug
+        * all_subnets  - supported with: txt, md, json, drawio, arch_drawio, svg, arch_svg, html, arch_html
```
adisos, line 14: why is this change related to this PR?

pkg/logging/logger.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Warnf writes a warning message to the log (unless DefaultLogger verbosity is set to LowVerbosity)
+func Warnf(format string, o ...interface{}) {
+	if Logger.verbosity >= MediumVerbosity {
+		pc, _, _, _ := runtime.Caller(1)
+		details := runtime.FuncForPC(pc)
+		Logger.l.Printf("WARN	%s	%s", details.Name(), fmt.Sprintf(format, o...))
+	}
+}
```
adisos, line 81: can you show examples for logger output with `details` for these functions?

README.md
olasaadi99: not related, it was not added to README by the one who added these lines to parse_args

adisos: oh. ok

pkg/logging/logger.go
adisos: I think it will be more readable without the details.. simply have:
```
WARN  ignoring floatingIP public-gw-ky: target.ResourceType public_gateway is not supported (only networkInterfaceResourceType supported)
```

pkg/logging/logger.go
```diff
@@ -0,0 +1,92 @@
...
+type DefaultLogger struct {
+	verbosity Verbosity
+	l         *log.Logger
+}
+
+// NewDefaultLogger creates an instance of DefaultLogger with the highest verbosity.
+func NewDefaultLogger() *DefaultLogger {
+	return NewDefaultLoggerWithVerbosity(HighVerbosity)
+}
+
```
zivnevo, line 38: Doesn't seem to be used

pkg/logging/logger.go
zivnevo, line 1: Please review all exported symbols and decide if they need to be exported

zivnevo (changes requested):

pkg/logging/logger.go
olasaadi99: what about returning error and debug , should I verbose these details?

zivnevo: I would leave details only for debug

olasaadi99: I think I would also leave details in return error (we still do not use it, will do next pr) it will print the error just when in high verbosity mode, used for debugging also

zivnevo (approved):
LGTM

pkg/logging/logger.go
adisos: I'd prefer to keep the returned errors as is, not through logger.. 
It will be difficult to maintain consistency across all project files, and through new development, to remember returning the error from this logger function...
With a simple grep one can find the error message, so no need to provide the exact function name.

adisos (approved):

PR #512, opened 2024-04-10
489 explain no tgw new

### LINKED ISSUE #489: bug in explainability multi-vpc when referring to src/dst with vpc name prefix

### LINKED ISSUE #500: explainability: specifying vsi address does not always work properly in multi-vpc context
Specifically, the following:
`go run .\cmd\analyzer -vpc-config C:\Users\SHIRIMORAN\npguard\vpc-network-config-analyzer\\pkg\ibmvpc\examples\input\input_multiVpc_larger_example_dup_names.json -analysis-type explain -src 10.240.3.5 -dst 10.240.12.4`
gets an error message:
`illegal src: internal address 10.240.3.5 not within any of the VPC's subnets' address range. exiting...`
Even though there is a vsi in that specific address

ShiriMoran:
replaces #502 
Fixes #489 
Fixes #500

### DIFF
--- a/pkg/ibmvpc/examples/out/explain_out/tgwEnabledSpecificFilter_all_vpcs_explain_debug.txt
+++ b/pkg/ibmvpc/examples/out/explain_out/tgwEnabledSpecificFilter_all_vpcs_explain_debug.txt
@@ -1,0 +1,30 @@
+Connectivity explanation between ky-vsi1-subnet20 and ky-vsi0-subnet2 within combined-vpc-local-tg-ky
+=====================================================================================================
+
+The following connection exists between ky-vsi1-subnet20[10.240.128.5] and ky-vsi0-subnet2[10.240.4.4]: All Connections
+
+Path:
+	ky-vsi1-subnet20[10.240.128.5] -> security group sg21-ky -> subnet20 -> network ACL acl21-ky -> 
+	test-vpc2-ky -> TGW local-tg-ky -> test-vpc0-ky -> 
+	network ACL acl2-ky -> subnet2 -> security group sg1-ky -> ky-vsi0-subnet2[10.240.4.4]
+
+
+Details:
+~~~~~~~~
+Egress:
+security group sg21-ky allows connection with the following allow rules
+	index: 1, direction: outbound, protocol: all, cidr: 0.0.0.0/0
+network ACL acl21-ky allows connection with the following allow rules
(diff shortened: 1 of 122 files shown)


ShiriMoran:
replaces #502

pkg/ibmvpc/examples/out/explain_out/tgwEnabledSpecificFilter_all_vpcs_explain_debug.txt
```diff
@@ -1,5 +1,5 @@
-Connectivity explanation between ky-vsi1-subnet20 and ky-vsi0-subnet2 within combined-vpc-local-tg-ky
```
adisos, line 1: missing the prefix vpc before the vsi name? ( e.g. `(vpc-name)/ky-vsi1-subnet20` ) ?

ShiriMoran: Its not "missing" since its not part of the name or of the input. In case of `multi-vpc` we can add the name of the relavant `vpc` for `src/dst` of the header.
Perhaps we can change the Name() function in case of multi-vpc to include the vpc?

adisos: ok, let's open another issue for this to be discussed and then decide.

ShiriMoran: #521

adisos (approved):

PR #543, opened 2024-04-18
Html improvments

haim-kermany:
1. set short label to connections
2. set full label to connections as toolTip
3. change doc header
4. fix node label overlapping each other
5. put explanation   as toolTip of destanation


pkg/drawio/connectivityMap.svg.tmpl
```diff
@@ -112,17 +119,18 @@
                 const selectableElems = document.querySelectorAll('[selectable="true"]');
                 var selectedElems = [];
                 const mainTitleText = 'Application connectivity graph'
-                const doFilterExplainText = 'For filtering, Double-click on a node/edge/legend item'
+                const doFilterExplainText = 'For filtering, Double-click on any item'
```
zivnevo, line 122:
```suggestion
                const doFilterExplainText = 'Double-click an item to filter-out unrelated entities'
```

pkg/drawio/connectivityMap.svg.tmpl
```diff
@@ -112,17 +119,18 @@
...
                 const unFilterExplainText = 'For unfiltering, Double-click on the background'
```
zivnevo, line 123:
```suggestion
                const unFilterExplainText = 'Double-click the background to cancel filtering'
```

pkg/drawio/connectivityMap.svg.tmpl
```diff
@@ -112,17 +119,18 @@
...
                 const selectSrcText = 'For connectivity explanation, Click the SOURCE node'
-                const selectDstText =  'Click the DESTINATION node'
+                const selectDstText =  'Click the DESTINATION node, And see destination toolTip '
                 const reselectSrcText = 'For another connectivity explanation, Click the SOURCE node'
 
                 var filterText = mainTitleText
                 var filterExplainText = doFilterExplainText
                 var explainText = ''
                 var explainExplainText = selectSrcText
+                var savesdTooltip = ''
```
zivnevo, line 133:
```suggestion
                var savedTooltip = ''
```

pkg/drawio/connectivityMap.svg.tmpl
```diff
@@ -201,11 +210,15 @@
 
                           // color the src and dst names
                           explainText = explainTitle + entry.textContent;
+                          toolTip = entry.textContent
                         }
                         else {
                           explainText = 'Did not find entry of <span style="background-color: yellow;">'+src_name+'</span>'+
                           ' and <span style="background-color: #ADD8E6;"> '+dst_name + '</span>';
+              						toolTip= 'Did not find entry of '+src_name+ '->' + dst_name
```
zivnevo, line 218: I prefer tooltips to never report errors.

pkg/drawio/connectivityMap.svg.tmpl
zivnevo, line 1: Why is it helpful to put the explanation as a tool tip on the destination?

zivnevo (changes requested):

pkg/drawio/createMapFile.go
```diff
@@ -105,6 +106,24 @@ func (data *templateData) SvgLabel(tn TreeNodeInterface) string {
...
+	maxConnLabelSize = 8
+	threeDots        = "..."
+)
+
+func (data *templateData) SvgShortLabel(tn TreeNodeInterface) string {
+	// the connection label is created in another package,
+	// so, instead of creating a short version, we edit the long version here:
+	label := data.SvgLabel(tn)
+	label = strings.ReplaceAll(label, "protocol:", "")
+	label = strings.ReplaceAll(label, "-ports:", "")
```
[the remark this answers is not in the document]
haim-kermany: got it
should I also omit the -ports?
i.e 
TCP src: 1-442,444-65535 dst: 443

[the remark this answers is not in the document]
zivnevo: I think not. Might be too confusing.
@adisos. @yorav, do you think otherwise?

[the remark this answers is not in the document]
kyorav: I think I agree with Ziv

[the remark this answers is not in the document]
haim-kermany: fixed

pkg/drawio/connectivityMap.svg.tmpl
haim-kermany: fixed

pkg/drawio/connectivityMap.svg.tmpl
haim-kermany: fixed

pkg/drawio/connectivityMap.svg.tmpl
haim-kermany: fixed

pkg/drawio/connectivityMap.svg.tmpl
haim-kermany: the idea is to avoid the scroll down every time you select a pair - the result is in your face.
yes, it is not intuitive. and it might confuse at the beginning.
it does not even work well (you need to move the mouse a little after selecting the destination) 
however, I think its a powerful feature

zivnevo (approved):

PR #550, opened 2024-04-25
rewriting getAll*

### LINKED ISSUE #356: implement and use getAll[Squares(),Icons(),Lines()]  in drawio
Consider a version of `getAllNodes()` which takes a filter function, and will not collect nodes for which the filter function returns `false`.

haim-kermany:
Fixes #356


pkg/drawio/treeNodeInterface.go
```diff
@@ -86,62 +88,77 @@ func setGeometry(tn TreeNodeInterface) {
...
-	for _, i := range icons {
-		childrenSet[i] = true
+// getSubTreeNodes() - return all the nodes in the sub tree
+
+func getSubTreeNodes(tn TreeNodeInterface) (
+	squares []SquareTreeNodeInterface,
+	icons []IconTreeNodeInterface,
+	lines []LineTreeNodeInterface) {
+	squares, icons, lines = tn.children()
+	for _, child := range joinTnsLists(squares, icons, lines) {
```
zivnevo, line 98:
```suggestion
	for _, child := range squares {
```
Icons and lines have no children. Better save recursive calls

pkg/drawio/treeNodeInterface.go
```diff
@@ -86,62 +88,77 @@ func setGeometry(tn TreeNodeInterface) {
...
-			childrenSet[s] = true
-		}
-	}
-	childrenSet[tn] = true
-	ret := []TreeNodeInterface{}
-	for c := range childrenSet {
-		ret = append(ret, c)
-	}
-	return ret
+	// remove duplications:
```
zivnevo, line 112: Why do we have duplicates?

zivnevo (changes requested):

pkg/drawio/treeNodeInterface.go
haim-kermany: icons can have lines as children (we want the line to be attached to routers if needed).
lines currently does not have children, but to be on the safe side, for further implementation, I prefer to assume that lines can have children

pkg/drawio/treeNodeInterface.go
haim-kermany: same icons can be in both a groupsquare, subnet, and/or security group

haim-kermany: fixed

pkg/drawio/treeNodeInterface.go
haim-kermany: yes, some are leaves.
but:
the return values for getSubTreeNodes() should contain the node itself. (if not, calling with root node will miss the root).
so, we must call getSubTreeNodes() for the leaves also.
however, a call with a leave will be thin.
i can add a stopping condition for leaves inside getSubTreeNodes(), but I dont think it will save a lot

zivnevo (approved):

PR #551, opened 2024-04-25
update analysis with new local field

olasaadi99:


pkg/ibmvpc/sg_analysis.go
```diff
@@ -317,6 +317,19 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
...
+	Locals := []*ipblock.IPBlock{}
+	for i := range rules {
+		if rules[i].local != nil {
+			Locals = append(Locals, rules[i].local)
+		}
+	}
+	disjointLocals := ipblock.DisjointIPBlocks(Locals, []*ipblock.IPBlock{ipblock.GetCidrAll()})
+	for i := range disjointLocals {
+		connectivityMap[disjointLocals[i]] = AnalyzeSGRules(rules, isIngress)
+	}
```
adisos, line 330: it seems that this will assign the same result to each disjoint local entry in the map, since each time ` AnalyzeSGRules` is called with the same set of rules, instead of considering only the relevant rules for this local entry

adisos (changes requested):

adisos:
It seems that the implementation is planned as my description below.
Please review and confirm.
I think it is important to add in this PR at least some basic unit tests that relate to these aspects and checks them. 

```
SG example:

members: n1, n2, n3 
targets: t1, t2, t3 

n1 in cidr1 only 
n2 in cidr2 only 
n3 in cidr1 union cidr2 

t1 in cidr1' only 
t2 in cidr2' only 
t3 in both cidr1' and cidr2' 


x,y,z: disjoint partition of cidr1,cidr2, such that x only in cidr1, y only in cidr2, z in intersection cidr1 and cidr2 

egress rules:
- local: cidr1, remote: cidr1' , conns: c1
- local: cidr2, remote: cidr2', conns: c2 


before local - one connectivity result: 
 dst x': c1
 dst y': c2
 dst z': c1 union c2
 
 
after local- connectivity result per disjoint local :
 
src z:
 dst x': c1
 dst y': c2
 dst z': c1 union c2
 
src y:
  dst cidr2': c2
  
src x:
  dst cidr1': c1
  
  
  
 AllowedConnectivity(n1, t1) - >  c1
 AllowedConnectivity(n1, t2) - >  none
 AllowedConnectivity(n1, t3) - >  c1
 
 AllowedConnectivity(n2, t1) - >  none
 AllowedConnectivity(n2, t2) - >  c2
 AllowedConnectivity(n2, t3) - >  c2
 
 
 AllowedConnectivity(n3, t1) - >  c1
 AllowedConnectivity(n3, t2) - >  c2
 AllowedConnectivity(n3, t3) - >  c1 union c2 
 
```

adisos:
Also, note that multiple disjoint locals entries could be having same connectivity results, so better caching the `AnalyzeSGRules` computations by set of rules being considered to this computation (to avoid re-computing identical results for different local key entries).

adisos:
Updated example description:

```
SG example:

cidr1 contained in cidr2
cidr1' contained in cidr2'

x,y,z: disjoint partition of cidr1,cidr2, such that x only in cidr2, y only in cidr2, z in intersection cidr1 and cidr2 ,: x < z < y 


members: n1, n2, n3 
targets: t1, t2, t3 

n1 in x
n2 in y
n3 in z

t1 in x'
t2 in y'
t3 in z'

egress rules:
- local: cidr1, remote: cidr1' , conns: c1
- local: cidr2, remote: cidr2', conns: c2 


before local - one connectivity result: 
 dst x': c2 
 dst z': c1 union c2
 dst y': c2
 
 
after local- connectivity result per disjoint local :

src x:
  dst cidr2': c2
 
src z:
 dst x': c2 
 dst z': c1 union c2
 dst y': c2
 
src y:
  dst cidr2': c2
  
  
  
 AllowedConnectivity(n1, t1) - >  c2
 AllowedConnectivity(n1, t2) - >  c2
 AllowedConnectivity(n1, t3) - >  c2
 
 AllowedConnectivity(n2, t1) - >  c2
 AllowedConnectivity(n2, t2) - >  c2
 AllowedConnectivity(n2, t3) - >  c2
 
 
 AllowedConnectivity(n3, t1) - >  c2
 AllowedConnectivity(n3, t2) - >  c2
 AllowedConnectivity(n3, t3) - >  c1 union c2 
 
```

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +110,75 @@ func TestSGRule(t *testing.T) {
...
+		"remote: 10.250.10.0-10.250.10.0, conn: protocol: TCP src-ports: 1-100 dst-ports: 5-1000\n"+
+		"remote: 10.250.10.1-10.250.10.1, conn: protocol: TCP src-ports: 1-100 dst-ports: 5-1000; "+
+		"protocol: UDP src-ports: 5-87 dst-ports: 10-3245\nremote: 10.250.10.2-10.250.10.3, conn: "+
+		"protocol: TCP src-ports: 1-100 dst-ports: 5-1000\nremote: 10.250.10.4-255.255.255.255, "+
+		"conn: No Connectionsremote: 0.0.0.0-10.250.9.255, conn: No Connections\nremote: 10.250.10.0-10.250.10.3,"+
+		" conn: protocol: TCP src-ports: 1-100 dst-ports: 5-1000\nremote: 10.250.10.4-255.255.255.255, conn: No Connections"+
+		"remote: 0.0.0.0-10.250.9.255, conn: No Connections\nremote: 10.250.10.0-10.250.10.3, "+
+		"conn: protocol: TCP src-ports: 1-100 dst-ports: 5-1000\nremote: 10.250.10.4-255.255.255.255, "+
+		"conn: No Connectionsremote: 0.0.0.0-255.255.255.255, conn: No Connectionsremote: "+
+		"0.0.0.0-255.255.255.255, conn: No Connections")
```
adisos, line 183: cannot follow the expected result in the current version..
also, better not use hard-coded strings for tests, as string formatting may change.. 
could be testing using `AllowedConnectivity` results? or through containment check of various connectivity entries in the map?

pkg/ibmvpc/sg_analysis.go
```diff
@@ -317,6 +317,26 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
...
+	disjointLocals := ipblock.DisjointIPBlocks(Locals, []*ipblock.IPBlock{ipblock.GetCidrAll()})
+	for i := range disjointLocals {
+		relevantRules := []*SGRule{}
+		for j := range rules {
+			if disjointLocals[i].ContainedIn(rules[j].local) {
+				relevantRules = append(relevantRules, rules[j])
+			}
+		}
+		// check if we already called AnalyzeSGRules with the same relevantRules
```
adisos, line 335: is this comment still a TODO? 
should also be handled.. (caching and reusing computed connectivity results for same set of rules) .
consider checking `pkg/common/genericSet.go` for generating key per set of rules .

adisos (changes requested):

pkg/ibmvpc/sg_analysis.go
adisos: also make sure there is a relevant test with caching scenario after this is added.

olasaadi99: The test I added has a cache scenario

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +110,101 @@ func TestSGRule(t *testing.T) {
...
+func TestNoLocalField(t *testing.T) {
+	cidr1t, _ := ipblock.FromCidrOrAddress("10.250.10.1")
+	cidr2t, _ := ipblock.FromCidrOrAddress("10.250.10.0/30")
+
+	c1 := connection.TCPorUDPConnection(netp.ProtocolString("UDP"), 5, 87, 10, 3245)
+	c2 := connection.TCPorUDPConnection(netp.ProtocolString("TCP"), 1, 100, 5, 1000)
+
+	rulesTest1 := []*SGRule{
+		{
+			target:      cidr1t,
```
adisos, line 123:
```suggestion
			target:      newIPBlockFromCIDROrAddressWithoutValidation("10.250.10.1"),
```
I think the test will be more readable this way, when IPBlock objects generated directly on test fields.
Check the other fields as well.

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +110,101 @@ func TestSGRule(t *testing.T) {
...
+	}
+
+	target1, _ := ipblock.FromCidrOrAddress("10.250.10.0")
+	target2, _ := ipblock.FromCidrOrAddress("10.250.10.2")
+	target3, _ := ipblock.FromCidrOrAddress("10.250.10.1")
+
+	egressConnectivityMap := make(map[*ipblock.IPBlock]*ConnectivityResult)
+	mapAndAnalyzeSGRules(rulesTest1, false, egressConnectivityMap)
+	require.Equal(t, len(egressConnectivityMap), 1)
+	for _, CR := range egressConnectivityMap {
```
adisos, line 143: rename `CR` to match golang naming conventions

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +110,101 @@ func TestSGRule(t *testing.T) {
...
+		},
+	}
+
+	target1, _ := ipblock.FromCidrOrAddress("10.250.10.0")
+	target2, _ := ipblock.FromCidrOrAddress("10.250.10.2")
+	target3, _ := ipblock.FromCidrOrAddress("10.250.10.1")
+
+	egressConnectivityMap := make(map[*ipblock.IPBlock]*ConnectivityResult)
+	mapAndAnalyzeSGRules(rulesTest1, false, egressConnectivityMap)
+	require.Equal(t, len(egressConnectivityMap), 1)
```
adisos, line 142:
```suggestion
	require.Equal(t, len(egressConnectivityMap), 1) //expecting size 1 because locals are only "0.0.0.0/0" in this example
```
explain `require` in documentation

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +110,101 @@ func TestSGRule(t *testing.T) {
...
+	for _, CR := range egressConnectivityMap {
+		for remote, conn := range CR.allowedConns {
+			if target1.ContainedIn(remote) || target2.ContainedIn(remote) {
+				require.True(t, conn.Equal(c2))
+			}
+			if target3.ContainedIn(remote) {
+				require.True(t, c1.ContainedIn(conn) && c2.ContainedIn(conn))
+			}
+		}
+	}
```
adisos, line 152: better be more explicit about expected content of  `CR.allowedConns` in the test, for example
can change to a switch over the set of expected remote `IPBlocks`, and for each case
verify that the connection is as expected. + also verify all switch cases were covered exactly once in the loop.
another option- assign an expected `connectivityResult` object , and use a method that compares equality of such objects. This can be used as table-driven tests for both tests TestNoLocalField and TestLocalField.

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +110,101 @@ func TestSGRule(t *testing.T) {
...
+			case member3.ContainedIn(local):
+				if target1.ContainedIn(remote) || target2.ContainedIn(remote) {
+					require.True(t, conn.Equal(c2))
+				}
+				if target3.ContainedIn(remote) {
+					require.True(t, c1.ContainedIn(conn) && c2.ContainedIn(conn))
+				}
+			}
+		}
+	}
```
adisos, line 209: same comment here.. unit tests should better be explicit, rather than code from which it is difficult to understand what properties are checked and why.

pkg/ibmvpc/sg_analysis.go
```diff
@@ -317,6 +318,34 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
 	return res
 }
 
+func mapAndAnalyzeSGRules(rules []*SGRule, isIngress bool, connectivityMap map[*ipblock.IPBlock]*ConnectivityResult) {
+	Locals := []*ipblock.IPBlock{}
```
adisos, line 322: rename `Locals` to `locals`

pkg/ibmvpc/sg_analysis.go
```diff
@@ -317,6 +318,34 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
...
+	for i := range rules {
+		if rules[i].local != nil {
+			Locals = append(Locals, rules[i].local)
+		}
+	}
+	disjointLocals := ipblock.DisjointIPBlocks(Locals, []*ipblock.IPBlock{ipblock.GetCidrAll()})
+	KeysToConnectivityResult := map[common.SetAsKey]*ConnectivityResult{}
```
adisos, line 329: rename `KeysToConnectivityResult` to  `keysToConnectivityResult`

pkg/ibmvpc/vpc.go
```diff
@@ -577,20 +577,22 @@ type SecurityGroup struct {
 
 func (sg *SecurityGroup) AllowedConnectivity(src, dst vpcmodel.Node, isIngress bool) *connection.Set {
 	memberStrAddress, targetIPBlock := sg.getMemberTargetStrAddress(src, dst, isIngress)
+	_, localIPBlock := sg.getMemberTargetStrAddress(src, dst, !isIngress)
```
adisos, line 580: better change the function `getMemberTargetStrAddress` to return `member.IPBlock()` as well, instead of calling it twice here

pkg/ibmvpc/sg_analysis.go
adisos: can you add a test to validate that caching actually worked on that scenario? currently that test could pass even if caching did not work from some reason.

adisos (changes requested):
added few more comments

pkg/ibmvpc/sg_analysis.go
```diff
@@ -16,18 +16,19 @@ import (
...
 	egressRules  []*SGRule
 	// rules are the default ones; that is, no rules were specified manually
-	isDefault           bool
-	ingressConnectivity *ConnectivityResult
-	egressConnectivity  *ConnectivityResult
-	sgMap               map[string]*SecurityGroup
-	referencedIPblocks  []*ipblock.IPBlock
+	isDefault              bool
+	ingressConnectivityMap map[*ipblock.IPBlock]*ConnectivityResult
+	egressConnectivityMap  map[*ipblock.IPBlock]*ConnectivityResult
```
adisos, line 29: consider a separate type for ` map[*ipblock.IPBlock]*ConnectivityResult`  (so it can be more documented what this represents, and  can added with methods, such as `equal` for tests)

adisos: could you simplify `ingressConnectivityMap` by only keeping the relevant entries that intersect with actual members of the SG ?  (no need to compute potential `ConnectivityResult` for members that do not exist in such range...)

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +111,215 @@ func TestSGRule(t *testing.T) {
...
+					fromIPRangeStrWithoutValidation("10.250.10.4-255.255.255.255"): {},
+					fromIPAddressStrWithoutValidation("10.250.10.1"):               {1, 2},
+				},
+				deniedConns: map[*ipblock.IPBlock]*connection.Set{},
+				denyRules:   map[*ipblock.IPBlock][]int{},
+			},
+		},
+	},
+	{
+		name: "localField",
```
adisos, line 173:
```suggestion
		name:  "local_field_in_rules_is_not_all_range", // a more interesting case of local field with effect on connectivity
```

pkg/ibmvpc/sg_analysis_test.go
```diff
@@ -107,3 +111,215 @@ func TestSGRule(t *testing.T) {
...
+	var connectivityResult1, connectivityResult2 *ConnectivityResult
+	for local, connectivityResult := range egressConnectivityMap {
+		if local.Equal(newIPBlockFromCIDROrAddressWithoutValidation("10.240.10.0")) {
+			connectivityResult1 = connectivityResult
+		}
+		if local.Equal(newIPBlockFromCIDROrAddressWithoutValidation("10.240.10.2/31")) {
+			connectivityResult2 = connectivityResult
+		}
+	}
+	require.True(t, connectivityResult1 == connectivityResult2)
```
adisos, line 324:
```suggestion
	require.True(t, connectivityResult1 == connectivityResult2) // // compare pointers-- to make sure that caching worked
```

pkg/ibmvpc/vpc.go
```diff
@@ -592,34 +592,34 @@ type SecurityGroup struct {
...
+	isIngress bool) (memberStrAddress, targetIPBlock *ipblock.IPBlock) {
 	var member, target vpcmodel.Node
 	if isIngress {
 		member, target = dst, src
 	} else {
 		member, target = src, dst
 	}
 	// TODO: member is expected to be internal node (validate?) [could use member.(vpcmodel.InternalNodeIntf).Address()]
-	return member.CidrOrAddress(), target.IPBlock()
+	return member.IPBlock(), target.IPBlock()
```
adisos, line 622: - add a third return value that preserves the original `member.CidrOrAddress()` value returned

pkg/ibmvpc/vpc.go
```diff
@@ -592,34 +592,34 @@ type SecurityGroup struct {
...
 		return false, nil, nil // connectivity not affected by this SG resource - input node is not its member
 	}
-	rules, err = sg.analyzer.rulesFilterInConnectivity(targetIPBlock, conn, isIngress)
+	rules, err = sg.analyzer.rulesFilterInConnectivity(targetIPBlock, memberIPBlock, conn, isIngress)
 	return true, rules, err
 }
 
 func (sg *SecurityGroup) getMemberTargetStrAddress(src, dst vpcmodel.Node,
-	isIngress bool) (memberStrAddress string, targetIPBlock *ipblock.IPBlock) {
+	isIngress bool) (memberStrAddress, targetIPBlock *ipblock.IPBlock) {
```
adisos, line 614:
```suggestion
	isIngress bool) (memberIPBlock, targetIPBlock *ipblock.IPBlock) {
```

pkg/ibmvpc/sg_analysis.go
olasaadi99: > could you simplify `ingressConnectivityMap` by only keeping the relevant entries that intersect with actual members of the SG ? (no need to compute potential `ConnectivityResult` for members that do not exist in such range...)

do not have access to the members

adisos: probably can be added from the parser

pkg/ibmvpc/sg_analysis.go
```diff
@@ -338,6 +338,21 @@ func mapAndAnalyzeSGRules(rules []*SGRule, isIngress bool, connectivityMap Conne
...
 	for i := range disjointLocals {
+		localContainsMember := false
+		for member := range currentSg.members {
+			memberIPBlock, err := ipblock.FromIPAddress(member)
+			if err != nil {
+				return
+			}
+			if memberIPBlock.ContainedIn(disjointLocals[i]) {
+				localContainsMember = true
+				break
```
adisos, line 349: can add a function (method for SecurityGroup) that returns an `*IPBlock` object with union of all members, have this value computed once outside the loop, and for each `disjointLocal` check that it has non-empty intersection with this value.

adisos (changes requested):

pkg/ibmvpc/connectivity_result.go
```diff
@@ -21,3 +26,96 @@ type ConnectivityResult struct {
...
+	// compare the concatenation result to validate equality of keys sets
+	for i := 0; i < len(keys1); i++ {
+		if keys1[i] != keys2[i] {
+			return false
+		}
+	}
+	for ip, indexes := range rules1 {
+		for otherIP, otherIndexes := range rules2 {
+			if ip.Equal(otherIP) {
+				if !reflect.DeepEqual(indexes, otherIndexes) {
```
adisos, line 86: are the indexes guaranteed to be in the same order ? or need to be sorted?

pkg/ibmvpc/sg_analysis.go
```diff
@@ -328,7 +328,7 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
 	return res
 }
 
-func mapAndAnalyzeSGRules(rules []*SGRule, isIngress bool, connectivityMap ConnectivityResultMap) {
+func mapAndAnalyzeSGRules(rules []*SGRule, isIngress bool, connectivityMap ConnectivityResultMap, currentSg *SecurityGroup) {
```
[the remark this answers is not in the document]
olasaadi99: yes it can, will update it

pkg/ibmvpc/sg_analysis.go
olasaadi99: I used this specific way because it can be non-empty intersection with this value but still we do not need this disjointLocal (no member alone contained in this disjointLocal)

pkg/ibmvpc/connectivity_result.go
olasaadi99: will sort before checking equality

pkg/ibmvpc/sg_analysis.go
adisos: I meant that each `disjointLocal` has to be checked with non empty intersection of the unified members ipBlock.
this means that at least one member is contained in this  `disjointLocal`.

pkg/ibmvpc/connectivity_result.go
```diff
@@ -21,3 +27,88 @@ type ConnectivityResult struct {
...
+		return false
+	}
+	keys1 := make([]string, len(first))
+	i := 0
+	for ipBlock := range first {
+		keys1[i] = ipBlock.String()
+		i += 1
+	}
+	i = 0
+	sort.Strings(keys1)
```
adisos, line 42: this part is also duplicated.. can be moved to a separate function that returns a list of sorted keys from such a map.

pkg/ibmvpc/vpc.go
```diff
@@ -608,34 +608,48 @@ type SecurityGroup struct {
...
+	memberIPBlock, targetIPBlock, memberStrAddress := sg.getMemberTargetStrAddress(src, dst, isIngress)
 	if _, ok := sg.members[memberStrAddress]; !ok {
 		return connection.None() // connectivity not affected by this SG resource - input node is not its member
 	}
-	return sg.analyzer.AllowedConnectivity(targetIPBlock, isIngress)
+	return sg.analyzer.AllowedConnectivity(targetIPBlock, memberIPBlock, isIngress)
+}
+
+// unifiedMembersIPBlock returns an *IPBlock object with union of all members IPBlock
+func (sg *SecurityGroup) UnifiedMembersIPBlock() (unifiedMembersIPBlock *ipblock.IPBlock, err error) {
```
adisos, line 619:
```suggestion
func (sg *SecurityGroup) unifiedMembersIPBlock() (unifiedMembersIPBlock *ipblock.IPBlock, err error) {
```

pkg/ibmvpc/vpc.go
```diff
@@ -608,34 +608,48 @@ type SecurityGroup struct {
...
+	unifiedMembersIPBlock = ipblock.New()
+	for member := range sg.members {
+		memberIPBlock, err := ipblock.FromIPAddress(member)
+		if err != nil {
+			return nil, err
+		}
+		unifiedMembersIPBlock = unifiedMembersIPBlock.Union(memberIPBlock)
```
adisos, line 626:
```suggestion
	for _, memberNode := range sg.members {		
		unifiedMembersIPBlock = unifiedMembersIPBlock.Union(memberNode.IPBlock())
```

pkg/ibmvpc/sg_analysis.go
```diff
@@ -327,6 +328,44 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
...
+	connectivityMap = make(ConnectivityResultMap)
+	locals := []*ipblock.IPBlock{}
+	for i := range rules {
+		if rules[i].local != nil {
+			locals = append(locals, rules[i].local)
+		}
+	}
+	disjointLocals := ipblock.DisjointIPBlocks(locals, []*ipblock.IPBlock{ipblock.GetCidrAll()})
+	keysToConnectivityResult := map[common.SetAsKey]*ConnectivityResult{}
+	unifiedMembersIPBlock, err := currentSg.UnifiedMembersIPBlock()
```
adisos, line 341:
```suggestion
	unifiedMembersIPBlock, err := currentSg.unifiedMembersIPBlock()
```

pkg/ibmvpc/sg_analysis.go
```diff
@@ -327,6 +328,44 @@ func AnalyzeSGRules(rules []*SGRule, isIngress bool) *ConnectivityResult {
...
+	if err != nil {
+		return nil
+	}
+	for i := range disjointLocals {
+		if disjointLocals[i].Intersect(unifiedMembersIPBlock).IsEmpty() {
+			continue
```
adisos, line 347:
```suggestion
			// no need to compute connectivity for local range that has no SG members within it
			continue
```

PR #559, opened 2024-05-01
521 vpc name

### LINKED ISSUE #498: connectivity output: subnet name should also include vpc , at least in the cross-vpc connectivity
open an issue with reference to this example output, where subnet name should also include vpc , at least in the cross-vpc connectivity ?

_Originally posted by @adisos in https://github.com/np-guard/vpc-network-config-analyzer/pull/495#discussion_r1549580467_

- [ ] cross-vpc output case to be handled first 
- [ ] consider handling duplicated names in the same vpc - see comment below

### LINKED ISSUE #521: In multivpc context - consider adding VPC to "name" of VSI and potentially more relevant resources
Its not "missing" since its not part of the Name() functionality or of the input. In case of `multi-vpc` I can add the name of the relevant `vpc` for `src/dst` of the header. What do you say @adisos ? I can add it in both headers; I'm not completely sure it is required, though. Given that the user is probably aware of what he queried about, and given that if there is ambiguity in input then we issue an error and ask the user to resolve it

_Originally posted by @ShiriMoran in https://github.com/np-guard/vpc-network-config-analyzer/pull/512#discussion_r1560696476_

ShiriMoran:
In multi-vpc context prefixes name with vpc name in explainability and connectivity

fixes #498 
fixes #521


pkg/vpcmodel/abstractVPC.go
```diff
@@ -40,10 +41,21 @@ type VPCResource struct {
 	VPCRef VPCResourceIntf `json:"-"`
 }
 
+func VPCPrefixMulti(c *VPCConfig, vpcName string) string {
+	if c.IsMultipleVPCsConfig {
+		return vpcName + "/"
+	}
+	return ""
+}
+
```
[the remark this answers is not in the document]
ShiriMoran: Unfortunately it won't work - the function `VPCPrefixMulti` is being called from concrete non-interface objects - `PrivateIP, IKSNode, NetworkInterface and ReservedIP` which can not call` func (n *VPCResource) ExtendedPrefix(c *VPCConfig)`

haim-kermany (approved):

pkg/ibmvpc/examples/out/explain_out/tgwAnotherExampleEnabledConn_all_vpcs_explain.txt
```diff
@@ -1,7 +1,7 @@
 Explaining connectivity from ky-vsi0-subnet5 to ky-vsi0-subnet11
 ================================================================
 
-Allowed connections from ky-vsi0-subnet5[10.240.9.4] to ky-vsi0-subnet11[10.240.80.4]: All Connections
+Allowed connections from test-vpc0-ky/ky-vsi0-subnet5[10.240.9.4] to test-vpc1-ky/ky-vsi0-subnet11[10.240.80.4]: All Connections
 
 Path:
 	ky-vsi0-subnet5[10.240.9.4] -> security group sg1-ky -> subnet5 -> network ACL acl3-ky ->
```
adisos, line 7: VSI name in the path should not be using `ExtendedName` format?

pkg/vpcmodel/abstractVPC.go
```diff
@@ -20,6 +20,8 @@ const (
 type VPCResourceIntf interface {
 	UID() string
 	Name() string
+	ExtendedName(*VPCConfig) string
+	ExtendedPrefix(config *VPCConfig) string
```
adisos, line 24: These methods better have some documentation

pkg/ibmvpc/vpc.go
```diff
@@ -57,7 +57,11 @@ func (r *ReservedIP) Name() string {
 	return getNodeName(r.vpe, r.Address())
 }
 
-// ReservedIP implements vpcmodel.Node interface
+func (r *ReservedIP) ExtendedName(c *vpcmodel.VPCConfig) string {
+	return r.ExtendedPrefix(c) + getNodeName(r.vpe, r.Address())
```
adisos, line 61:
```suggestion
	return r.ExtendedPrefix(c) + r.Name()
```
avoid duplicating the code implemented within `Name()` method

pkg/ibmvpc/vpc.go
```diff
@@ -68,6 +72,10 @@ func (pip *PrivateIP) Name() string {
 	return getNodeName(pip.loadBalancer, pip.Address())
 }
 
+func (pip *PrivateIP) ExtendedName(c *vpcmodel.VPCConfig) string {
+	return pip.ExtendedPrefix(c) + getNodeName(pip.loadBalancer, pip.Address())
+}
+
```
adisos, line 78:
```suggestion
func (pip *PrivateIP) ExtendedName(c *vpcmodel.VPCConfig) string {
	return pip.ExtendedPrefix(c) + pip.Name()
}

```

pkg/ibmvpc/vpc.go
```diff
@@ -83,6 +91,10 @@ func (ni *NetworkInterface) Name() string {
 	return getNodeName(ni.vsi, ni.Address())
 }
 
+func (ni *NetworkInterface) ExtendedName(c *vpcmodel.VPCConfig) string {
+	return ni.ExtendedPrefix(c) + getNodeName(ni.vsi, ni.Address())
+}
```
adisos, line 96:
```suggestion
func (ni *NetworkInterface) ExtendedName(c *vpcmodel.VPCConfig) string {
	return ni.ExtendedPrefix(c) + ni.Name()
}
```

pkg/ibmvpc/vpc.go
```diff
@@ -97,6 +109,10 @@ func (n *IKSNode) Name() string {
 	return getNodeName(n.ResourceName, n.Address())
 }
 
+func (n *IKSNode) ExtendedName(c *vpcmodel.VPCConfig) string {
+	return n.ExtendedPrefix(c) + getNodeName(n.ResourceName, n.Address())
+}
```
adisos, line 114:
```suggestion
func (n *IKSNode) ExtendedName(c *vpcmodel.VPCConfig) string {
	return n.ExtendedPrefix(c) + n.Name()
}
```

pkg/ibmvpc/vpc.go
```diff
@@ -319,6 +335,10 @@ func (nl *NaclLayer) Name() string {
 	return ""
 }
 
+func (nl *NaclLayer) ExtendedName(config *vpcmodel.VPCConfig) string {
+	return ""
+}
+
```
adisos, line 341:
```suggestion
```
seems this method is not needed?

pkg/ibmvpc/vpc.go
```diff
@@ -493,6 +513,10 @@ func (sgl *SecurityGroupLayer) Name() string {
 	return ""
 }
 
+func (sgl *SecurityGroupLayer) ExtendedName(config *vpcmodel.VPCConfig) string {
+	return ""
+}
+
```
adisos, line 519:
```suggestion
```
same here

pkg/vpcmodel/abstractVPC.go
```diff
@@ -40,10 +42,21 @@ type VPCResource struct {
...
+}
+
 func (n *VPCResource) Name() string {
 	return n.ResourceName
 }
 
+func (n *VPCResource) ExtendedName(c *VPCConfig) string {
+	return n.ExtendedPrefix(c) + n.Name()
+}
+
```
adisos, line 59: why is `ExtendedName` defined in the level of `VPCResource` ?  seems it is only used in the level of `Node` and of `EndpointElem`  and Subnet ?
could be added only to `InternalNode` and `NodeSet` interfaces?

adisos (changes requested):

pkg/ibmvpc/examples/out/explain_out/tgwAnotherExampleEnabledConn_all_vpcs_explain.txt
ShiriMoran: No, in the path I believe they should not be using the `ExtendedName`  format. The VPC is part of the path anyways. I find it clearer this way.

pkg/vpcmodel/abstractVPC.go
ShiriMoran: done

pkg/vpcmodel/abstractVPC.go
ShiriMoran: For consistency, I defined it in the same level as the one Name() is defined as.
Moreover, `ExtendedName(..)` is called for objects that can potentially be `ExternalNetwork`. Defining `ExternalName` only for `InternalNode` and `NodeSet` interfaces will require checking in each such place what is the underlying type of the object and then calling either `Name()` or `ExtendedName(...)` (seems the best way of doing it is defining another member function of `VPCResourceIntf`). So all together we will have to maintain more (and I believe messier) code than just defining `ExtendedName(..)` for `ExternalNetwork`.

pkg/ibmvpc/vpc.go
ShiriMoran: I think it is, see answer to the last comment

pkg/ibmvpc/vpc.go
ShiriMoran: also here, I think this is the best option. Lets discuss in the last comment

pkg/vpcmodel/abstractVPC.go
adisos: but isn't `ExtendedName` relevant only for endpoints in multi-vpc context? (and not any vpc resource?)
seems if it's defined in the level of `Node` and `NodeSet` it may be sufficient and not require conversions?

ShiriMoran: Yes, this should work. Just note that in this case ExtendedName will still need to be defined for ExternalNetwork.  Code lines will hardy be reduced; but I guess it makes more sense design wise.

ShiriMoran: No, it does not work since `grouping` and `jsonOutput` uses `VPCResourceIntf` interface which can be either `Node` or `EndpointElem`
This change will require another abstract type - I doubt it is worth it. 

I removed the implementation of `ExtendedName()` from `nacl` and `sg` where it was not really needed, though.

pkg/vpcmodel/abstractVPC.go
```diff
@@ -20,6 +20,10 @@ const (
 type VPCResourceIntf interface {
 	UID() string
 	Name() string
+	// ExtendedName Name, name prefixed with vpc in multi-vpc context, name otherwise
```
adisos, line 23:
```suggestion
	// ExtendedName returns a resource name that includes its VPC as prefix when necessary.	
	// for example, a subnet with name "s1" within VPC "v1" will have extended name: "v1/s1"
	// note this method is relevant only for Node and Subnet objects.
	// note it adds the prefix only for input config that has multiple VPCs context. 
```

pkg/vpcmodel/abstractVPC.go
```diff
@@ -20,6 +20,10 @@ const (
...
+	ExtendedName(*VPCConfig) string
+	// ExtendedPrefix vpc name with deliminator in multi-vpc context, empty string otherwise
+	ExtendedPrefix(config *VPCConfig) string
```
adisos, line 26:
```suggestion
	// ExtendedPrefix returns the prefix to be added for ExtendedName, given the input config 
	ExtendedPrefix(config *VPCConfig) string
```

adisos (approved):

PR #564, opened 2024-05-02
521 rule with sg

### LINKED ISSUE #447: Improve  printing of SG Rules for explainability analysis
The original rule syntax does not enable multiple CIDRs per rule, I assume `cidr: 10.240.10.4/32,10.240.30.5/32,10.240.30.6/32` is the result of an original rule that allows some SG members.. should consider (in a separate issue) to present the original rule, and not the interpretation to CIDRs in this case.

_Originally posted by @adisos in https://github.com/np-guard/vpc-network-config-analyzer/pull/435#discussion_r1520177203_

ShiriMoran:
fixes #447


haim-kermany (approved):

pkg/ibmvpc/sg_analysis.go
```diff
@@ -272,11 +277,16 @@ func (sga *SGAnalyzer) getSGrules(sgObj *vpc1.SecurityGroup) (ingressRules, egre
...
+	sgName string // target specified is SG
+}
+
 type SGRule struct {
-	target      *ipblock.IPBlock
+	remote      ruleTarget
 	connections *connection.Set
 	index       int // index of original rule in *vpc1.SecurityGroup.Rules
-	local       *ipblock.IPBlock
+	local       ruleTarget
```
adisos, line 289: local can remain of type `IPBlock`, it cannot reference another SG

adisos (changes requested):

pkg/ibmvpc/sg_analysis.go
ShiriMoran: right. restored

adisos: forgot to push ?

ShiriMoran: No, weird. It was pushed.

adisos (approved):

PR #571, opened 2024-05-07
consistency in the return values of VPC.AddressRange() and VPC.AddressPrefixes()

olasaadi99:
change return value of VPC.AddressPrefixes() to return *ipblock.IPBlock


ShiriMoran (changes requested):
Please add tests; otherwise the functionality is not used; we need some indication it works

ShiriMoran (approved):

PR #575, opened 2024-05-09
Multi sg in drawio

haim-kermany:
add support to multi sg in drawio.

main changes: 
1. an icon has a list of SGs that it is in.
2. PartialSGTreeNode can now represent more than one SG.
3. the layout code that handled an SG, is now it is handling a list of icons that has the same list of SGs


pkg/drawio/squareTreeNode.go
```diff
@@ -198,30 +198,44 @@ func (tn *SGTreeNode) children() ([]SquareTreeNodeInterface, []IconTreeNodeInter
...
+	labels := []string{}
+	for _, sg := range tn.sgs {
+		labels = append(labels, sg.name)
+	}
+	return labels
 }
+
 func (tn *PartialSGTreeNode) Kind() string {
-	return tn.parent.Kind()
+	kind := tn.sgs[0].Kind()
```
zivnevo, line 234: This doesn't seem to work well. In the html I see `Connectivity graph of s:sg1,sg3`

pkg/drawio/squareTreeNode.go
```diff
@@ -198,30 +198,44 @@ func (tn *SGTreeNode) children() ([]SquareTreeNodeInterface, []IconTreeNodeInter
...
+	}
 	return &psg
 }
 func (tn *PartialSGTreeNode) children() ([]SquareTreeNodeInterface, []IconTreeNodeInterface, []LineTreeNodeInterface) {
 	return nil, tn.elements, tn.connections
 }
-func (tn *PartialSGTreeNode) DrawioParent() TreeNodeInterface {
-	return tn.Parent().Parent()
+func (tn *PartialSGTreeNode) labels() []string {
+	labels := []string{}
```
zivnevo, line 226: preallocate memory

pkg/drawio/layout.go
```diff
@@ -470,6 +469,26 @@ func (ly *layoutS) setSGLocations() {
 	}
 }
 
+// sortIconsBySGs() sort all the icons by their SGs
+// return a list of lists of icons - all the icons in one list have the same sg
```
zivnevo, line 473:
```suggestion
// return a list of lists of icons - all the icons in one list have exactly the same set of sgs
```

pkg/drawio/layout.go
```diff
@@ -470,6 +469,26 @@ func (ly *layoutS) setSGLocations() {
...
+	// remove duplicates
+	icons = common.FromList(icons).AsList()
+	// get the icons list for every group of sgs:
+	sgsToIcons := map[setAsKey][]TreeNodeInterface{}
+	for _, icon := range icons {
+		sgsAsKey := icon.SGs().AsKey()
+		sgsToIcons[sgsAsKey] = append(sgsToIcons[sgsAsKey], icon)
+	}
+	// covert to list:
+	return common.MapValues(sgsToIcons)
```
zivnevo, line 489: Why do we need this extra conversion? Why not return a map?

zivnevo (changes requested):

pkg/drawio/squareTreeNode.go
haim-kermany: in drawio_test I do not call 	SetKind(string), so its empty.
you can see results at 
pkg/ibmvpc/examples/out/analysis_out/out_iks_workers_large_all_vpcs__with_grouping.html

pkg/drawio/squareTreeNode.go
haim-kermany: fixed

pkg/drawio/layout.go
haim-kermany: fixed

pkg/drawio/layout.go
haim-kermany: the keys in the map has now meaning.
the desired output of the method should be a list of list.
I think its better coding, thought we pay small run time

pkg/drawio/squareTreeNode.go
zivnevo: Got it. It would be nicer if we don't use abbreviations in `Kind()`.
That is: better have `Connectivity graph of Security Group` rather than `Connectivity graph of SG`.
Maybe open another issue for this one.

zivnevo (approved):

PR #585, opened 2024-05-19
handle empty squares on subnet mode

haim-kermany:
fixing 
https://github.com/np-guard/vpc-network-config-analyzer/issues/583
current approach:
subnet grouping worked on subnet groups, and ignore subnets that was not part of a group
it cause 2 bugs:
1. squares that does not have zones did not have locations.
2. zones with subnets that are not in groups was pushed to the end of canvas (which might be in another cloud).

the new approach take care of all squares


pkg/drawio/subnetsLayout.go
zivnevo, line 1: Add a 5th step to the description of the high-level algorithm at the top of the file:
```
// 5. layout subnets with no groups and empty zones/regions/vpcs/ ...
```

zivnevo (approved):

PR #587, opened 2024-05-20
Using cobra to parse command-line args - phase 2

zivnevo:
In this PR:
1. Moving all the code for building VPC-configs and calling the analyzer to `subcmds` packages (this leaves a very slim `main`)
2. The subcommands now call this code directly from their `RunE` method.
3. Root command takes care of initializing the logger, in its `PersistentPreRun` method
4. The `InArgs` struct is no longer exported
5. Package documentation (in `root.go`)

Command-line syntax is left unchanged.


adisos (approved):
Looks good!
LGTM

PR #588, opened 2024-05-20
add "load blalncer" prefix to reports

haim-kermany:


pkg/ibmvpc/vpc.go
```diff
@@ -244,6 +244,10 @@ type LoadBalancer struct {
 	listeners []LoadBalancerListener
 }
 
+func (lb *LoadBalancer) ExtendedName(c *vpcmodel.VPCConfig) string {
+	return lb.Kind() + ":" + lb.ExtendedPrefix(c) + lb.Name()
+}
+
```
adisos, line 250: - the function can be `func (lb *LoadBalancer) Name()` instead of `func (lb *LoadBalancer) ExtendedName()` ?
- I would change the format to (example): `service-alb[LoadBalancer]`

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -656,7 +656,7 @@ func TestAllWithComparison(t *testing.T) {
 		if formatsAvoidComparison[tt.format] {
 			tt.mode = outputIgnore
 		} else {
-			tt.mode = outputComparison
+			tt.mode = outputGeneration
```
adisos, line 659: why is this changed?

adisos (changes requested):

pkg/ibmvpc/analysis_output_test.go
haim-kermany: sorry, fixed

pkg/ibmvpc/vpc.go
haim-kermany: -all resource has both Name() and ExtendedName() (please notice that ExtendedName() call Nam()). in Drawio We use Name() (we need it there to be short). ExtendedName() is for textual reports.

- fixed

adisos: This is the current documentation for `ExtendedName` 
```
	// ExtendedName returns a resource name that includes its VPC as prefix when necessary.
	// for example, a subnet with name "s1" within VPC "v1" will have extended name: "v1/s1"
	// note this method is relevant only for Node and Subnet objects.
	// note it adds the prefix only for input config that has multiple VPCs context.
	ExtendedName(*VPCConfig) string
```

So this function for LB breaks this semantics..

in textual reports both `Name` and `ExtendedName` are used .
how about having a new method for VPCResource interface `ShortName` , to be used for drawio? a method that only returns the resource name ?

haim-kermany: From the code its look like the textual report always uses ExtendedName(*VPCConfig), however, if we do not in multiple VPCs context, the ExtendedPrefix() returns ""
I updated documentation

ShortName() is not needed, for LB in drawio,  I just use the resourceName

adisos: opened issue #590

adisos (approved):

PR #597, opened 2024-05-23
Generic

elazarg:
This PR demonstrates the changes required in order to use the generic data structures (and ICMP set) in https://github.com/np-guard/models/pull/33. 95% of the changes are renaming, but some are related to the introduction of a dedicated ICMP set.

Some tests still do not pass, due to an issue with tracking statefulness (which IMO should be part of this repo and not np-guard/models).

The line `replace github.com/np-guard/models => ../models` in go.mod is useful for local experimentation until https://github.com/np-guard/models/pull/33 is merged.


adisos (commented):
@ShiriMoran please check if this PR can be changed to have all planned changes for stateful analysis (removing files `protocols_aux_test.go` and `statefulness.go`)

elazarg:
Updated to accommodate #634. 
The tests cannot pass in github. Locally:
* Some tests are updated to explicitly state the only code in ICMP types that have only single code.
* Two tests fail due to what seems like an inconsistent ordering of partitions. It might be a good idea to use ordered maps instead of hash maps.

elazarg:
Okay, now all the tests pass. The issue was the use of CartesianRightTriple instead of CartesianLeftTriple when partitioning TCPUDPSet. The decision is currently at the library side, but it should probably move to the client side.

Issue #199, opened 2023-10-17
Struct with full stateful information


ShiriMoran:
Dependency:  #255 
Currently  `func (conn *ConnectionSet) Subtract(other *ConnectionSet) *ConnectionSet` ignores the `statefullness` aspect
There are two cases here:
1. is the delta connection stateful
2. `connectionProperties` is identical but `conn stateful`  is not

The 1st item can not be computed easily since the `stateful`  property is not fine grained.
There are two ways to solve it:
1. Fine grain connection as per the stateful property; this implies that if one protocol is stateful for a certain `<src,dst>` and the other is not then they will not reside in the same connection
2. Compute the statefullness after computing the Delta for each item in the Delta

We agreed as a 1st stage to solve #170 without the stateful aspect

adisos:
For reference:
The place where currently connections are not yet split accurately by stateful property, but rather assigned value `false` if any subset is not stateful: https://github.com/np-guard/vpc-network-config-analyzer/blob/main/pkg/vpcmodel/nodesConnectivity.go#L242-L246

The place where such more accurate split occurs (in the json output format):  https://github.com/np-guard/vpc-network-config-analyzer/blob/main/pkg/vpcmodel/jsonOutput.go#L27-L40

adisos:
May be relevant to other operations as well (such as `union`)

ShiriMoran:
Some connection arithmetic has an ERROR return value while other does not. This inconsistent behavior is not desired. Look into it.

ShiriMoran:
#255 solution might result in this issue being redundant

ShiriMoran:
*Solution* add the following struct:
```
// ExtendedSet connection details
type ExtendedSet struct {
	statefulConn    *connection.Set // stateful TCP connection between <src, dst>
	nonStatefulConn *connection.Set // nonstateful TCP connection between <src, dst>; complementary of statefulConn
	otherConn       *connection.Set // non TCP connection (for which stateful is non-relevant)
	conn            *connection.Set // entire connection
}
```
The result of connectivity analysis (across the broad) will be `ExtendedSet`; `IsStateful` will be now removed from `*Set`
`ExtendedSet` is defined as follows:
* `statefulConn`: `stateful TCP` connection 
* `otherConn: non-TCP` connection
* `conn`: the entire connection
* `nonStatefulConn`: `TCP` minus `statefulConn`
[...]

PR #606, opened 2024-06-03
255 stateful connectivity refactor

### LINKED ISSUE #255: A single ConnectionSet should represent either all statefull or all non-statefull connections
At the moment a single connection may contain stateful and non-stateful connections and a connection is considered stateful if it has at least one non-stateful connection.
Better to separate such a connection into stateful and non-stateful lines, as a result of which the connection description and potential arithmetic will be more accurate.
Follow up #199

ShiriMoran:
addresses #255 and #199
Complete solution includes changes in models/connections; due to mutual dependencies between the changes the entire solution will have the following stage (this stage is the biggest one)
1.  **vpcAnalyzer**: define `ExtendedSet` that is built on `connection.Set` and stores the accurate stateful info and refactor the following to use the new struct

- connectivity endpoints/subnets analyzer
- semantic diff 
- explainability 
2. **models/connectivity**

- remove the no longer used `IsStateful` from `connection.Set`. 
- prepare `models/connectivity` package to extracting the stateful functionality into  `vpcAnalyzer`; this mainly includes exporting functionality that is used by the stateful functionality

3. **vpcAnalyzer**: copy (with minor refactoring) stateful functionality and unitests from models
4. **models/connectivity** delete stateful related functionality and tests


haim-kermany:
please use generics

ShiriMoran:
> please use generics

Please open an issue, we'll do it in a separate PR. Its not as trivial and out of the (accurate) scope of this PR.

haim-kermany:
please overview the operators,
equal() need a fix
isAll() - revisit the use?

ShiriMoran:
> please overview the operators, equal() need a fix isAll() - revisit the use?

Fixed. Good catch!

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,114 @@ SPDX-License-Identifier: Apache-2.0
...
-import "github.com/np-guard/models/pkg/connection"
+import (
+	"github.com/np-guard/models/pkg/connection"
+	"github.com/np-guard/models/pkg/netp"
+)
+
+// todo: remove stateful from connection.Set (for both options)
+
+// ExtendedSet connection details
+type ExtendedSet struct {
```
haim-kermany, line 17: consider rename "ExtendedSet"
its more than a Set..

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,114 @@ SPDX-License-Identifier: Apache-2.0
...
+func NoConnsExtendedSet() *ExtendedSet {
+	return &ExtendedSet{
+		statefulConn:    NoConns(),
+		nonStatefulConn: NoConns(),
+		otherConn:       NoConns(),
+		conn:            NoConns(),
+	}
+}
+
+var all = connection.All()
```
haim-kermany, line 38: since its a global variable, rename all to something more specific, or remove it

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,114 @@ SPDX-License-Identifier: Apache-2.0
...
+	return e.statefulConn.Equal(other.statefulConn) && e.otherConn.Equal(other.otherConn) &&
+		e.conn.Equal(other.conn)
+}
+
+func (e *ExtendedSet) Copy() *ExtendedSet {
+	return &ExtendedSet{
+		statefulConn:    e.statefulConn,
+		nonStatefulConn: e.nonStatefulConn,
+		otherConn:       e.otherConn,
+		conn:            e.conn,
```
haim-kermany, line 58: you can just copy the pointers, you need deeper copy

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,114 @@ SPDX-License-Identifier: Apache-2.0
...
+func (e *ExtendedSet) Intersect(other *ExtendedSet) *ExtendedSet {
+	statefulConn := e.statefulConn.Intersect(other.statefulConn)
+	otherConn := e.otherConn.Intersect(other.otherConn)
+	conn := e.conn.Intersect(other.conn)
+	return &ExtendedSet{
+		statefulConn:    statefulConn,
+		nonStatefulConn: computeNonStatefulConn(conn, otherConn, statefulConn),
+		otherConn:       otherConn,
+		conn:            conn,
+	}
```
haim-kermany, line 75: its look like there are some repeating patterns of creating ExtendedSet.
I would consider defining a constructor for each pattern.
for example, here the constructor will get statefulConn, otherConn and conn, only.
computeNonStatefulConn() should be called inside the constructor

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -17,15 +124,38 @@ func (connectivityMap GeneralConnectivityMap) updateAllowedConnsMap(src, dst VPC
...
+func newTCPSet() *connection.Set {
+	return connection.TCPorUDPConnection(netp.ProtocolStringTCP, connection.MinPort, connection.MaxPort,
+		connection.MinPort, connection.MaxPort)
+}
+
+func partitionTCPNonTCP(conn *connection.Set) (tcp, nonTCP *connection.Set) {
+	tcpFractionOfConn := newTCPSet().Intersect(conn)
+	nonTCPFractionOfConn := conn.Subtract(tcpFractionOfConn)
+	return tcpFractionOfConn, nonTCPFractionOfConn
+}
```
haim-kermany, line 161: partitionTCPNonTCP() should be called in another constructor

haim-kermany (changes requested):

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: renamed to SetWithStateful

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: removed

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: fixed

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: right. done.

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: Not sure I follow; construction to what? the returned objects here are Set. Anyways `partitionTCPNonTCP()` will not stay here after the next PR; it will be moved to models.

pkg/vpcmodel/commonConnectivity.go
haim-kermany: its not a set, the name should be something that related to connections

ShiriMoran: `ConnWithStateful`

pkg/vpcmodel/commonConnectivity.go
haim-kermany: for now, you have two constuctors:
1. NewStateWithStateful() (need to be renamed)
2. EmptyConnWithStateful()
there are place for more constractors:
3. &ConnWithStateful{statefulConn: newTCPSet(), nonStatefulConn: NoConns(),
		otherConn: connection.All().Subtract(newTCPSet()), conn: connection.All()}
4. 				tcpFraction, nonTCPFraction := partitionTCPNonTCP(conn)
				v.AllowedConnsCombinedStateful.updateAllowedStatefulConnsMap(src, dst,
					&ConnWithStateful{statefulConn: tcpFraction, otherConn: nonTCPFraction,
						nonStatefulConn: connection.None(), conn: conn})
5. maybe more...
6.

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,97 @@ SPDX-License-Identifier: Apache-2.0
...
+// todo: remove stateful from connection.Set (for both options)
+
+// ConnWithStateful connection details
+type ConnWithStateful struct {
+	statefulConn    *connection.Set // stateful TCP connection between <src, dst>
+	nonStatefulConn *connection.Set // nonstateful TCP connection between <src, dst>; complementary of statefulConn
+	otherConn       *connection.Set // non TCP connection (for which stateful is non-relevant)
+	conn            *connection.Set // entire connection
+}
+
```
haim-kermany, line 23:
```suggestion
type Connection struct {
	statefulTCP   *connection.Set // stateful TCP connection between <src, dst>
	nonStatefulTCP *connection.Set // nonstateful TCP connection between <src, dst>; complementary of statefulConn
	nonTCP       *connection.Set // non TCP connection (for which stateful is non-relevant)
	all            *connection.Set // entire connection
}

```

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: 1 renamed
3 and 4 - code refactored to use the existing constructor (I don't see the point in having a constructor for a single use; moreover, the existing constructor is good enough in the relevant places)

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: Added `NewConnWithStatefulGivenTCPStatefulAndNonTCP`

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,139 @@ SPDX-License-Identifier: Apache-2.0
...
+// operation on ConnWithStateful
+// The operations are performed on the disjoint statefulConn and otherConn and on allConn which contains them;
+// nonStatefulConn - the tcp complementary of statefulConn w.r.t. allConn -
+// is computed as allConn minus (statefulConn union otherConn)
+
+func computeNonStatefulConn(allConn, otherConn, statefulConn *connection.Set) *connection.Set {
+	return allConn.Subtract(otherConn).Subtract(statefulConn)
+}
+
+func EmptyConnWithStateful() *ConnWithStateful {
```
adisos, line 33: add documentation for exported function, here and in the rest of the places later

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,139 @@ SPDX-License-Identifier: Apache-2.0
...
+func (e *ConnWithStateful) String() string {
+	return e.allConn.String()
+}
+
+func (e *ConnWithStateful) EnhancedString() string {
+	if !e.nonStatefulConn.IsEmpty() {
+		return e.String() + " *"
+	}
+	return e.String()
+}
```
adisos, line 136: Shouldn't rename `EnhancedString()` to be `String()`, and remove the simple `String()` method? 
If one is looking for connection string without consideration of statefuleness, he can just use `e.allConn.String()`

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,139 @@ SPDX-License-Identifier: Apache-2.0
...
-import "github.com/np-guard/models/pkg/connection"
+import (
+	"github.com/np-guard/models/pkg/connection"
+	"github.com/np-guard/models/pkg/netp"
+)
+
+// todo: remove stateful from connection.Set (for both options)
+
+// ConnWithStateful connection details
+type ConnWithStateful struct {
```
adisos, line 17: move this new type and its methods/functions to a separate file, such as `conn_with_stateful.go` (golang file names conventions are with underscores).

pkg/vpcmodel/vpcConnectivity.go
```diff
@@ -23,20 +23,13 @@ type VPCConnectivity struct {
...
-	AllowedConnsCombined GeneralConnectivityMap
-
 	// allowed connectivity combined and stateful
 	// used by debug and json format only (at the moment)
 	// For src node provides a map of dsts and the stateful connection it has to these dsts
 	// note that subset of a non-stateful connection from AllowedConnsCombined can still be stateful
 	// and as such add to this map
-	AllowedConnsCombinedStateful GeneralConnectivityMap
+
+	AllowedConnsCombinedStateful GeneralStatefulConnectivityMap
```
adisos, line 32: please open an issue to consider enhancing current connectivity reports with more accurate statefulness info, as available at this new type. (as opposed to only json that was adding this so far) .

adisos (changes requested):
few initial comments, did not review all files yet.

pkg/ibmvpc/examples/out/analysis_out/lb_bad_practice_all_vpcs_.txt
```diff
@@ -4,7 +4,7 @@ alb[LoadBalancer] => vsi0-sub1[10.240.0.4] : All Connections**
 alb[LoadBalancer] => vsi0-sub2[10.240.64.4] : All Connections**
 alb[LoadBalancer] => vsi0-sub3[10.240.128.4] : All Connections**
 vsi0-ctrl-sub[10.240.2.4] => Public Internet (all ranges) : All Connections
-vsi0-ctrl-sub[10.240.2.4] => alb[LoadBalancer] : All Connections**
+vsi0-ctrl-sub[10.240.2.4] => alb[LoadBalancer] : All Connections ***
```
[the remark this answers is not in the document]
ShiriMoran: its a combination of the below. I agree it is confusing and should be fixed, but not in this PR. This PR revealed this because the LB abstraction prior to this PR ignores the stateful property. 
We should probably open an issue for V0.5, the same release and the other relevant PRs that causes this ***
```

connections are stateful (on TCP) unless marked with *
connections marked with ** are an over-approximation, not all private IPs have the same connectivity
```

pkg/vpcmodel/vpcConnectivity.go
ShiriMoran: you mean two issues - one for connectivity and one for diffs. (We can prioritize each)

pkg/ibmvpc/examples/out/analysis_out/lb_bad_practice_all_vpcs_.txt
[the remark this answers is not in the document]
adisos: at least add a space after the first `*` for now?

pkg/vpcmodel/vpcConnectivity.go
adisos: or two tasks in the same issue

ShiriMoran: #615 and #614 
I've put them in separate issue so that they could be assigned to e.g. different milestone

pkg/ibmvpc/examples/out/analysis_out/lb_bad_practice_all_vpcs_.txt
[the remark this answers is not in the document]
ShiriMoran: done, we should consider further improving - in which case we should open a followup issue

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,139 @@ SPDX-License-Identifier: Apache-2.0
...
 
-import "github.com/np-guard/models/pkg/connection"
+import (
+	"github.com/np-guard/models/pkg/connection"
+	"github.com/np-guard/models/pkg/netp"
+)
+
+// todo: remove stateful from connection.Set (for both options)
+
+// ConnWithStateful connection details
```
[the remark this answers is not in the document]
ShiriMoran: It need not be exported, but I'll improve it as you suggest

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: I'll first go over the functions - see which really needs to be exported

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,139 @@ SPDX-License-Identifier: Apache-2.0
...
+	statefulConn    *connection.Set // stateful TCP connection between <src, dst>
+	nonStatefulConn *connection.Set // nonstateful TCP connection between <src, dst>; complementary of statefulConn
+	otherConn       *connection.Set // non TCP connection (for which stateful is non-relevant)
+	allConn         *connection.Set // entire connection
+}
+
+// operation on ConnWithStateful
+// The operations are performed on the disjoint statefulConn and otherConn and on allConn which contains them;
+// nonStatefulConn - the tcp complementary of statefulConn w.r.t. allConn -
+// is computed as allConn minus (statefulConn union otherConn)
```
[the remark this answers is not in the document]
ShiriMoran: Because it is used for determine whether the connection is stateful or not. In fact, it is the property that is accessed the most. So better not compute it each time.

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: No. There are cases String() is used - where we want to print the connection without the * for stateful. E.g. in explainability and in the debug report

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: done

pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -59,7 +59,8 @@ func (c *VPCConfig) GetVPCNetworkConnectivity(grouping, lbAbstraction bool) (res
 		}
 	}
 	res.computeAllowedConnsCombined()
```
haim-kermany, line 61: I think this call is not needed

pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -177,7 +178,8 @@ func switchSrcDstNodes(switchOrder bool, src, dst Node) (srcRes, dstRes Node) {
 	return src, dst
 }
 
-func (v *VPCConnectivity) computeCombinedConnectionsPerDirection(isIngressDirection bool, node Node, connectivityRes *ConnectivityResult) {
+func (v *VPCConnectivity) computeCombinedConnectionsPerDirection(isIngressDirection bool, node Node,
+	connectivityRes *ConnectivityResult, allowedConnsCombined GeneralConnectivityMap) {
```
haim-kermany, line 182: Its not C++
do not put the result (allowedConnsCombined) as an input

pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -188,19 +190,19 @@ func (v *VPCConnectivity) computeCombinedConnectionsPerDirection(isIngressDirect
...
-	v.AllowedConnsCombined = GeneralConnectivityMap{}
+// the stateful state of the connectivity is not computed here
+func (v *VPCConnectivity) computeAllowedConnsCombined() GeneralConnectivityMap {
+	allowedConnsCombined := GeneralConnectivityMap{}
 	for node, connectivityRes := range v.AllowedConns {
-		v.computeCombinedConnectionsPerDirection(true, node, connectivityRes)
-		v.computeCombinedConnectionsPerDirection(false, node, connectivityRes)
+		v.computeCombinedConnectionsPerDirection(true, node, connectivityRes, allowedConnsCombined)
+		v.computeCombinedConnectionsPerDirection(false, node, connectivityRes, allowedConnsCombined)
 	}
```
haim-kermany, line 204: create GeneralConnectivityMap{} inside computeCombinedConnectionsPerDirection(), return it as an output, and use
updateMap() to merge the two outputs

pkg/vpcmodel/nodesConnectivity.go
ShiriMoran: you are right. removed.

pkg/vpcmodel/nodesConnectivity.go
ShiriMoran: Not sure its a good idea. In the worst case it will double the complexity as each edge will be accessed twice: once for computing it and once for adding it.

pkg/vpcmodel/nodesConnectivity.go
ShiriMoran: The output (and input) is now the receiver

pkg/vpcmodel/commonConnectivity.go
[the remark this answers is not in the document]
adisos: some methods are still exported, though shouldn't be (such as `Subtract`)

pkg/vpcmodel/commonConnectivity.go
adisos: I see you renamed and moved to `detaildConn.go`  -the golang file name convention is with underscore, so better use `detailed_conn.go` (for example see https://blog.emka.web.id/2022/09/project-filename-convention-in-golang.html) 

also, the previous version was `connWithStateful` which is more specific than `detailedConn` - why renaming?

pkg/vpcmodel/commonConnectivity.go
adisos: I still find id confusing.. shouldn't the main method `String()` include stateful information? 
and have a dedicated method such as `ConnWithoutStateful()` ?

pkg/vpcmodel/commonConnectivity.go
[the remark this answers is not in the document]
adisos: then I would add a comment in the struct type, that this field should not be assigned directly, and is always computed from the other 3 fields, when building such object.

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,7 +6,139 @@ SPDX-License-Identifier: Apache-2.0
...
+
+// todo: remove stateful from connection.Set (for both options)
+
+// ConnWithStateful connection details
+type ConnWithStateful struct {
+	statefulConn    *connection.Set // stateful TCP connection between <src, dst>
+	nonStatefulConn *connection.Set // nonstateful TCP connection between <src, dst>; complementary of statefulConn
+	otherConn       *connection.Set // non TCP connection (for which stateful is non-relevant)
+	allConn         *connection.Set // entire connection
+}
```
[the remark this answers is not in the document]
adisos: why not? shouldn't it be defined for the abstraction as well?
maybe let's have a discussion..

cmd/analyzer/main_test.go
adisos, line 1: why is this filed changed?

pkg/vpcmodel/commonConnectivity.go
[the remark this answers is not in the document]
ShiriMoran: I've unexported all the operations; only Equal which is used (commented) by the ibmvpc package is exported

pkg/vpcmodel/commonConnectivity.go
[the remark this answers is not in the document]
ShiriMoran: ok, lets discuss.

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: done

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: 1. I followed our named convention - e.g. `externalNetwork, debugOutput, nodesConnectivity`  and more. If I'll change it as you suggest we will have unconsistency in the package, which I find more troubling. If we are not consistent with go's naming convention, best to have a separate issue and rename all relevant files.
2. I renamed as result of Haim's CR; I agreed with him it is better not to have with as part of the name. Maybe `connStatefulDetail`? a bit too long? `connStateful` is a bit misleading I think

pkg/vpcmodel/commonConnectivity.go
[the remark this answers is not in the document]
ShiriMoran: Hope it is clear enough now

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: In 2nd though, I've deleted `string()` and used `String()` of `allConns` in the places it was used. Then I've renamed `enhancedString()` to `string()`

pkg/vpcmodel/commonConnectivity.go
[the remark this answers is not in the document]
adisos: opened issue #617

pkg/vpcmodel/commonConnectivity.go
adisos: opened issue #618

cmd/analyzer/main_test.go
ShiriMoran: because the test failed - even though I never touched it; fails seems unconnected to my PR. 
Was never supposed to be merged.
I will change it back, lets see if it keeps failing.

pkg/vpcmodel/detailedConn.go
```diff
@@ -0,0 +1,113 @@
...
+	tcpRspDisable *connection.Set // non-responsive TCP connection between <src, dst>; complementary of tcpRspEnable
+	// connection is defined to be responsive if nonTCP is empty
+}
+
+// operation on detailedConn
+// The operations are performed on the disjoint tcpRspEnable and nonTCP and on allConn which contains them;
+// tcpRspDisable - the tcp complementary of tcpRspEnable w.r.t. allConn -
+// is computed as allConn minus (tcpRspEnable union nonTCP)
+
+func newDetailConn(tspRspConn, otherConn, allConn *connection.Set) *detailedConn {
```
adisos, line 34:
```suggestion
func newDetailedConn(tspRspConn, otherConn, allConn *connection.Set) *detailedConn {
```

pkg/vpcmodel/detailedConn.go
```diff
@@ -0,0 +1,113 @@
...
+	return &detailedConn{
+		tcpRspEnable:  tspRspConn,
+		tcpRspDisable: (allConn.Subtract(otherConn)).Subtract(tspRspConn),
+		nonTCP:        otherConn,
+		allConn:       allConn,
+	}
+}
+
+func emptyDetailConn() *detailedConn {
```
adisos, line 43:
```suggestion
func emptyDetailedConn() *detailedConn {
```

pkg/vpcmodel/detailedConn.go
```diff
@@ -0,0 +1,113 @@
...
+	return newDetailConn(NoConns(), NoConns(), NoConns())
```
adisos, line 44:
```suggestion
	return newDetailedConn(NoConns(), NoConns(), NoConns())
```

and in the rest of the functions here

pkg/vpcmodel/commonConnectivity.go
```diff
@@ -6,26 +6,40 @@ SPDX-License-Identifier: Apache-2.0
...
 
-import "github.com/np-guard/models/pkg/connection"
+import (
+	"github.com/np-guard/models/pkg/connection"
+)
+
+// GeneralResponsiveConnectivityMap describes connectivity
+type GeneralResponsiveConnectivityMap map[VPCResourceIntf]map[VPCResourceIntf]*detailedConn
 
 type GeneralConnectivityMap map[VPCResourceIntf]map[VPCResourceIntf]*connection.Set
```
adisos, line 17: update documentation here (for `GeneralResponsiveConnectivityMap` and `GeneralConnectivityMap`)

pkg/vpcmodel/detailedConn.go
```diff
@@ -0,0 +1,113 @@
...
+// it is kept since it is widely used - to determine if the connection is responsive
+type detailedConn struct {
+	tcpRspEnable  *connection.Set // responsive TCP connection between <src, dst>
+	nonTCP        *connection.Set // non TCP connection (for which responsiveness is non-relevant)
+	allConn       *connection.Set // entire connection
+	tcpRspDisable *connection.Set // non-responsive TCP connection between <src, dst>; complementary of tcpRspEnable
+	// connection is defined to be responsive if nonTCP is empty
+}
+
+// operation on detailedConn
```
adisos, line 29:
```suggestion
// operations on detailedConn
```

adisos (commented):
few more documentation comments

pkg/vpcmodel/detailedConn.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// subtract of two detailedConn: subtraction of tcpRspEnable, nonTCP and allConn
+// (tcpRspDisable is computed based on these)
+func (e *detailedConn) subtract(other *detailedConn) *detailedConn {
+	rspConn := e.tcpRspEnable.Subtract(other.tcpRspEnable)
+	otherConn := e.nonTCP.Subtract(other.nonTCP)
+	conn := e.allConn.Subtract(other.allConn)
+	return newDetailConn(rspConn, otherConn, conn)
+}
```
adisos, line 91: subtract is not necessarily required, as in the abstraction we just check if there is a missing connection, thus equality check and bool result should be sufficient.

haim-kermany: 1. in general, I think that even though subtract is not really needed, its a good idea to implement it now, by someone who expert on this, and not leave the implementation to someone who will need it in the feature.
2. specific to the abstraction, I would like to report abstraction information information in the explainability somehow

cmd/analyzer/main_test.go
ShiriMoran: fixed

pkg/vpcmodel/detailedConn.go
ShiriMoran: Changed
(I thought it would be better to have a shorter name)

pkg/vpcmodel/detailedConn.go
ShiriMoran: done

pkg/vpcmodel/commonConnectivity.go
ShiriMoran: done

pkg/vpcmodel/detailedConn.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/np-guard/models/pkg/netp"
+)
+
+// todo: remove stateful from connection.Set
+
+// detailedConn captures the connection with connection's responsive details, as described below.
+// It is created from src to dest allowed connection (TCP and non-TCP) and response dest to src allowed connection
+// (TCP and non-TCP); further entities of the connection may be created from operations as union e.g. for abstraction
+// note: tcpRspDisable is not independent and is calculated based on the other properties;
+// it is kept since it is widely used - to determine if the connection is responsive
```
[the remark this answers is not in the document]
ShiriMoran: done (I clarified some of the text)

pkg/vpcmodel/detailedConn.go
ShiriMoran: changed in relevant places

haim-kermany (approved):

adisos (approved):

PR #612, opened 2024-06-06
Avoid printing usage message on semantic errors

### LINKED ISSUE #611: usage printed on any error
The usage instructions should be printed only when there is an error in the command-line syntax. Attached is an example of an unsupported config object and the command-line and printout looks like this:
```
$ ../../vpc-network-config-analyzer/bin/vpcanalyzer report subnets -o=txt -c=config_object.json
Error: unsupported connectivity map with partial subnet ranges per connectivity result
Usage:
  vpcanalyzer report subnets [flags]

Flags:
  -h, --help   help for subnets

Global Flags:
      --debug                    runs in debug mode
      --dump-resources string    file path to store resources collected from the cloud provider
  -o, --format string            output format; must be one of [json, txt, md, drawio, arch_drawio, svg, arch_svg, html, arch_html, debug] (default "")
  -g, --grouping                 whether to group together endpoints sharing the same connectivity
      --output-file string       file path to store results
  -p, --provider string          collect resources from an account in this cloud provider
  -q, --quiet                    runs quietly, reports only severe errors and results
  -r, --region stringArray       cloud region from which to collect resources, can pass multiple regions
      --resource-group string    resource group id or name from which to collect resources
  -v, --verbose                  runs with more informative messages printed to log
      --vpc string               CRN of the VPC to analyze
  -c, --vpc-config stringArray   file paths to input configs, can pass multiple config files

2024/06/05 16:17:18 unsupported connectivity map with partial subnet ranges per connectivity result. exiting...
```

The error should be printed only once, and the usage should not be printed at all in this case.
[config_object.json](https://github.com/user-attachments/files/15604558/config_object.json)

zivnevo:
Fixes #611

### DIFF
--- a/cmd/analyzer/main.go
+++ b/cmd/analyzer/main.go
@@ -7,10 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package main
 
 import (
-	"log"
 	"os"
 
 	"github.com/np-guard/vpc-network-config-analyzer/cmd/analyzer/subcmds"
+	"github.com/np-guard/vpc-network-config-analyzer/pkg/logging"
 )
 
 // The actual main function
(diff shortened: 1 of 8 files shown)


adisos (approved):
looks good!

PR #631, opened 2024-06-17
LB Explainability in HTML

haim-kermany:


pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,6 +29,10 @@ type rulesConnection struct {
 	egressRules  rulesInLayers
 }
 
+// loadBalancerDennyEgressRule - is a rule applied on load balancer private IP:
+// a private IP can have egress connection to only to a pool member of the load balancer
+type loadBalancerDennyEgressRule bool
```
zivnevo, line 34:
```suggestion
type loadBalancerDenyEgressRule bool
```

pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -86,6 +91,11 @@ func explainMissingCrossVpcRouter(src, dst string, connQuery *connection.Set) st
 		"connect them", noConnectionHeader(src, dst, connQuery)+newLine)
 }
 
+func explainLoadBalancerDenyEgress(src, dst string, connQuery *connection.Set) string {
+	return fmt.Sprintf("%v\nConnection from a private IP is allowed only to load balancer pool members\n",
```
zivnevo, line 95:
```suggestion
	return fmt.Sprintf("%v\nPrivate IPs of a load balancer can only connect to its pool members\n",
```

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,6 +29,10 @@ type rulesConnection struct {
 	egressRules  rulesInLayers
 }
 
+// loadBalancerDennyEgressRule - is a rule applied on load balancer private IP:
+// a private IP can have egress connection to only to a pool member of the load balancer
```
zivnevo, line 33:
```suggestion
// loadBalancerDennyEgressRule - is a rule applied to all private IPs of a given load balancer:
// these private IPs can only connect to pool members of the load balancer
```

zivnevo (changes requested):
A few minor things

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityPrint.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -456,15 +472,15 @@ func (details *rulesAndConnDetails) computeConnections(c *VPCConfig,
 // if src or dst is a node then the node is from getCidrExternalNodes,
 // thus there is a node in VPCConfig that either equal to or contains it.
 func (v *VPCConnectivity) getConnection(c *VPCConfig, src, dst Node) (conn *detailedConn, err error) {
-	srcForConnection, err1 := c.getContainingConfigNode(src)
+	srcForConnection, err1 := c.getConnectedResource(src)
```
ShiriMoran, line 475: A pip is translated/abstracted to its LB?

pkg/vpcmodel/multiExplainability.go
```diff
@@ -88,6 +88,8 @@ func (c *VPCConfig) getNodesFromEndpoint(endpoint EndpointElem) ([]Node, error)
 	switch n := endpoint.(type) {
```
ShiriMoran, line 88: Somehow this file is not very familiar to me... Lets take it offline.

ShiriMoran (changes requested):

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -44,6 +48,8 @@ type srcDstDetails struct {
 	crossVpcRules  []RulesInTable  // cross vpc (only tgw at the moment) prefix rules effecting the connection (or lack of)
 	// there could be more than one connection effecting the connection since src/dst cidr's may contain more than one AP
 
+	// loadBalancerDenyEgressRule - true if src is pip, and dst is not pool member:
+	loadBalancerRule loadBalancerDenyEgressRule
```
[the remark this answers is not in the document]
haim-kermany: fixed

pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -69,7 +69,12 @@ func (explanation *Explanation) String(verbose bool) string {
 	linesStr := make([]string, len(explanation.groupedLines))
 	groupedLines := explanation.groupedLines
 	for i, groupedLine := range groupedLines {
-		linesStr[i] += groupedLine.explainabilityLineStr(explanation.c, explanation.connQuery, verbose) +
+		if groupedLine.commonProperties.expDetails.loadBalancerRule {
```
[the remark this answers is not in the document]
haim-kermany: fixed

pkg/vpcmodel/multiExplainability.go
haim-kermany: done

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: yes

pkg/ibmvpc/vpc.go
```diff
@@ -85,6 +85,14 @@ func (pip *PrivateIP) ExtendedName(c *vpcmodel.VPCConfig) string {
 	return pip.ExtendedPrefix(c) + pip.Name()
 }
 
+// AbstractedNodeSet returns the pip load balancer if it was abstracted
+func (pip *PrivateIP) AbstractedNodeSet() vpcmodel.NodeSet {
```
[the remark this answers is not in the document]
haim-kermany: changed to AbstractedToNodeSet(), though, I do not think that AbstractedNodeSet() is better.
if you can think of a better name, please suggest

ShiriMoran (commented):
Please add a test demonstrating the explanation you added

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -48,8 +69,8 @@ type srcDstDetails struct {
 	crossVpcRules  []RulesInTable  // cross vpc (only tgw at the moment) prefix rules effecting the connection (or lack of)
 	// there could be more than one connection effecting the connection since src/dst cidr's may contain more than one AP
 
-	// loadBalancerDenyEgressRule - true if src is pip, and dst is not pool member:
-	loadBalancerRule loadBalancerDenyEgressRule
+	// loadBalancerRule - the lb rule affecting this connection.
```
ShiriMoran, line 72: please explain shortly in what case this rule is relevant and why it is unique - namely, why there can not be more than one LB rule relevant

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,9 +29,30 @@ type rulesConnection struct {
...
 
-// loadBalancerDennyEgressRule - is a rule applied to all private IPs of a given load balancer:
+// the load balancer is a rule applied to all private IPs of a given load balancer:
 // these private IPs can only connect to pool members of the load balancer
-type loadBalancerDenyEgressRule bool
+type LoadBalancerRule struct {
+	// the relevant load balancer:
+	lb LoadBalancer
+	//	Denny- true if src is pip, and dst is not pool member:
+	denny bool
```
ShiriMoran: Why not have such a rule only when it is relevant - deny is true? Why this additional bool property?

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,9 +29,30 @@ type rulesConnection struct {
 	egressRules  rulesInLayers
 }
 
-// loadBalancerDennyEgressRule - is a rule applied to all private IPs of a given load balancer:
+// the load balancer is a rule applied to all private IPs of a given load balancer:
 // these private IPs can only connect to pool members of the load balancer
-type loadBalancerDenyEgressRule bool
+type LoadBalancerRule struct {
```
ShiriMoran, line 34: Is this specific to IBM?
If it is, wouldn't it make sense to implement in the VPC package? and keep here only some abstraction?

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,9 +29,30 @@ type rulesConnection struct {
...
+	lb LoadBalancer
+	//	Denny- true if src is pip, and dst is not pool member:
+	denny bool
+}
+
+func NewLoadBalancerRule(lb LoadBalancer, denny bool) *LoadBalancerRule {
+	return &LoadBalancerRule{lb, denny}
+}
+func (lbr *LoadBalancerRule) Denny() bool { return lbr.denny }
+func (lbr *LoadBalancerRule) StringDetailsOfRule() string {
```
ShiriMoran, line 45: This should surely not be here. 
If it is not in the IBM package then at least put in in `explainabilityPrint.go`

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,9 +29,30 @@ type rulesConnection struct {
...
+	return lbr.StringHeaderOfRule() + " to destinations which are its pool members\n"
```
ShiriMoran, line 46: Why is this separation to `StringHeaderOfRule()` and `StringDetailsOfRule()`  required?

ShiriMoran (commented):
Few initial comments (I will be back ;-) )

pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -114,16 +114,22 @@ func (g *groupedConnLine) explainabilityLineStr(c *VPCConfig, connQuery *connect
...
 	egressBlocking := !expDetails.egressEnabled && needEgress
-	var externalRouterHeader, crossRouterFilterHeader, resourceEffectHeader,
-		crossRouterFilterDetails, details string
+	var externalRouterHeader, crossRouterFilterHeader, loadBalancerHeader, resourceEffectHeader,
+		crossRouterFilterDetails, loadBalancerDetails, details string
 	externalRouter, crossVpcRouter, crossVpcRules := expDetails.externalRouter, expDetails.crossVpcRouter, expDetails.crossVpcRules
 	if externalRouter != nil && (src.IsExternal() || dst.IsExternal()) {
 		externalRouterHeader = "External traffic via " + externalRouter.Kind() + ": " + externalRouter.Name() + newLine
 	}
+	if loadBalancerRule != nil {
```
ShiriMoran, line 129: In the other cases the header prints only whether the filter/cross vpc router enables the connection or blocks it and the details prints the details of the specific rule in the table/prefix filter that effects the connection. These details includes the index of the rule and e.g. its target, direction (inbound/outbound) etc.

However, here we don't really have such a rule. And the only relevant details are already contained in the header. So I'm not sure this separation is required here. The same text can be used for both

pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -185,14 +191,24 @@ func (g *groupedConnLine) explainPerCaseStr(c *VPCConfig, src, dst EndpointElem,
...
+		if ingressBlocking {
+			blockedBy = append(blockedBy, "ingress")
+		}
+		if egressBlocking {
+			blockedBy = append(blockedBy, "egress")
+		}
+		l := len(blockedBy)
+		blockedByString := fmt.Sprintf("by %s", blockedBy[0])
+		if l > 1 {
+			blockedByString = fmt.Sprintf("both by %s and %s",
```
ShiriMoran, line 208: If it is blocked by all 3 then both is not good...

ShiriMoran (changes requested):
Some more comments.

pkg/vpcmodel/explainabilityPrint.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityPrint.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: Load balancer is not an IBM concept.
currently, we have LoadBalancer as an interface at vpcmodel
and LoadBalancer as a struct in the ibmvpc.

I guess that LoadBalancerRule should be an interface too and implemented differently on each vendor,  but I do not have all the information.
does load balancer from other vendors effect on the connectivity, I assume yes, but how?
there is also high chance that all LoadBalancerRules from all vendors will share the same behavior, and there will be no need for interface.

for now, I wrapped  all in functions, to make it easy to change it to an interface later

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -172,16 +196,18 @@ func (c *VPCConfig) computeExplainRules(srcNodes, dstNodes []Node,
...
-// 4. The actual relevant filter, depending on whether src xor dst is external
+// 4. The external filters relevant to the <src, dst> given the external routingResource
+// 5. The internal filters relevant to the <src, dst>
+// 6. The actual relevant filter, depending on whether src xor dst is external
 func (details *rulesAndConnDetails) computeRoutersAndFilters(c *VPCConfig) (err error) {
 	for _, singleSrcDstDetails := range *details {
 		// RoutingResources are computed by the parser for []Nodes of the VPC,
 		src := singleSrcDstDetails.src
 		dst := singleSrcDstDetails.dst
+		singleSrcDstDetails.loadBalancerRule = c.getLoadBalancerRule(src, dst)
```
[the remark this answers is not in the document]
haim-kermany: yes

pkg/vpcmodel/abstractVPC.go
```diff
@@ -110,6 +110,8 @@ type Node interface {
 	// IsPublicInternet returns true if the node is external,
 	// currently nodes which are external but not public Internet are ignored
 	IsPublicInternet() bool
+	// AbstractedToNodeSet returns the abstracted nodeSet that contains this node (if any)
```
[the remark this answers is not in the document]
haim-kermany: fixed

adisos (commented):
Missing example tests and output examples in this PR? 
The textual output format is also impacted for explainability command in this PR? (from changes in `explainabilityPrint.go` ?)

pkg/vpcmodel/explainabilityConnectivity.go
```diff
@@ -29,6 +29,28 @@ type rulesConnection struct {
...
+	return &LoadBalancerRule{lb, deny}
+}
+func (lbr *LoadBalancerRule) Deny() bool { return lbr.deny }
+
+func (lbr *LoadBalancerRule) String() string {
+	action := "allow"
+	if lbr.Deny() {
+		action = "blocks"
+	}
+	return fmt.Sprintf("load balancer %s %s connection to destinations which are its pool members\n", lbr.lb.Name(), action)
```
ShiriMoran, line 51: Instead of
`load balancer kube-clusterid:1-3a95ce10fa514d48ab490f9eda1a6db1[LoadBalancer] allow connection to destinations which are its pool members`
Please write
`load balancer kube-clusterid:1-3a95ce10fa514d48ab490f9eda1a6db1[LoadBalancer] allows connection to <xyz> since it is one of its pool members`
Similarly:
In 
`Load Balancer: load balancer kube-clusterid:1-3a95ce10fa514d48ab490f9eda1a6db1[LoadBalancer] blocks connection to destinations which are its pool members `
1. No need for double Load balancer in the prefix
2. ".... blocks connection to <css> since its not one of its pool members"
Similarly in the detail of blocks and in the header of allows

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: decided to make it an interface, with implementation in the ibmvpc.

pkg/vpcmodel/explainabilityConnectivity.go
haim-kermany: fixed

haim-kermany:
> Missing example tests and output examples in this PR? The textual output format is also impacted for explainability command in this PR? (from changes in `explainabilityPrint.go` ?)

this PR effect only the html files. so test will be added in the next PR
the explainabilityPrint.go is backward compatiable

pkg/ibmvpc/vpc.go
```diff
@@ -321,6 +321,30 @@ func (lb *LoadBalancer) AbstractionInfo() *vpcmodel.AbstractionInfo {
...
+	return &LoadBalancerRule{lb, deny, src, dst}
+}
+
+func (lbr *LoadBalancerRule) Deny() bool { return lbr.deny }
+
+func (lbr *LoadBalancerRule) String() string {
+	if lbr.Deny() {
+		return fmt.Sprintf("%s blocks connection to %s since it is not one of its pool members\n", lbr.lb.Name(), lbr.dst.Name())
+	}
+	return fmt.Sprintf("%s allow connection to %s since it is one of its pool members\n", lbr.lb.Name(), lbr.dst.Name())
```
ShiriMoran, line 345:
```suggestion
	return fmt.Sprintf("%s allows connection to %s which is one of its pool members\n", lbr.lb.Name(), lbr.dst.Name())
```

pkg/ibmvpc/vpc.go
```diff
@@ -321,6 +321,30 @@ func (lb *LoadBalancer) AbstractionInfo() *vpcmodel.AbstractionInfo {
...
+
+func NewLoadBalancerRule(lb *LoadBalancer, deny bool, src, dst vpcmodel.Node) vpcmodel.LoadBalancerRule {
+	return &LoadBalancerRule{lb, deny, src, dst}
+}
+
+func (lbr *LoadBalancerRule) Deny() bool { return lbr.deny }
+
+func (lbr *LoadBalancerRule) String() string {
+	if lbr.Deny() {
+		return fmt.Sprintf("%s blocks connection to %s since it is not one of its pool members\n", lbr.lb.Name(), lbr.dst.Name())
```
ShiriMoran, line 343:
```suggestion
		return fmt.Sprintf("%s blocks connection to %s which is not one of its pool members\n", lbr.lb.Name(), lbr.dst.Name())
```

ShiriMoran (approved):

adisos (approved):

zivnevo (approved):

PR #659, opened 2024-07-01
sunset debug flag

haim-kermany:


adisos:
- please link the PR to a relevant issue
- please add description of the PR -- how you handle the old debug flag

pkg/ibmvpc/parser.go
```diff
@@ -155,9 +155,7 @@ func VPCConfigsFromResources(rc *datamodel.ResourcesContainerModel, vpcID, resou
 		return nil, err
 	}
 
-	if debug {
-		printVPCConfigs(res)
-	}
+	printVPCConfigs(res)
```
zivnevo, line 158: We are making here tons of calls to `logging.Debugf()` which will do nothing if verbosity is not set to high. How about adding a method to the logger to get its verbosity level, and only call `printVPCConfigs()` if the verbosity level is high?

zivnevo (changes requested):

pkg/ibmvpc/parser.go
haim-kermany: fixed

zivnevo (approved):

adisos (approved):

PR #671, opened 2024-07-03
Lb expalinability in cli

haim-kermany:
expose the LB explainability to the cli.
1. user can give a LB name as a query input.
2. user can not give an address of a private IP as an input. when a range addresses is given, the pip is filter out


pkg/ibmvpc/examples/out/analysis_out/filters_split_lb_subnet_all_vpcs__no_lbAbstract.txt
```diff
@@ -36,70 +15,91 @@ alb[LB private IP][10.240.65.4] => vsi0-sub3[10.240.128.4] : All Connections
 alb[LB private IP][10.240.65.4] => vsi1-sub1[10.240.0.4] : All Connections
 alb[LB private IP][10.240.65.4] => vsi1-sub2[10.240.64.4] : All Connections
 alb[LB private IP][10.240.65.4] => vsi1-sub3[10.240.128.5] : All Connections
+alb[Potential LB private IP][10.240.1.0/25] => vsi0-sub1[10.240.0.5] : All Connections
```
ShiriMoran, line 18: The src pip are not the same as before.
E.g.: 10.240.1.4 before vs 10.240.1.0/25
Is this intended?
And why is it?
How is it related to adding LB to the cli?

pkg/ibmvpc/examples/out/analysis_out/filters_split_lb_subnet_all_vpcs__no_lbAbstract.txt
```diff
@@ -36,70 +15,91 @@ alb[LB private IP][10.240.65.4] => vsi0-sub3[10.240.128.4] : All Connections
...
 vsi0-sub3[10.240.128.4] => vsi0-sub2[10.240.64.5] : All Connections
 vsi0-sub3[10.240.128.4] => vsi1-sub1[10.240.0.4] : All Connections
 vsi0-sub3[10.240.128.4] => vsi1-sub2[10.240.64.4] : All Connections
 vsi0-sub3[10.240.128.4] => vsi1-sub3[10.240.128.5] : All Connections
-vsi1-sub1[10.240.0.4] => alb[Fake LB private IP][10.240.1.128] : All Connections
-vsi1-sub1[10.240.0.4] => alb[Fake LB private IP][10.240.1.4] : All Connections
-vsi1-sub1[10.240.0.4] => alb[Fake LB private IP][10.240.129.128] : All Connections
-vsi1-sub1[10.240.0.4] => alb[Fake LB private IP][10.240.65.128] : All Connections
 vsi1-sub1[10.240.0.4] => alb[LB private IP][10.240.65.4] : All Connections
+vsi1-sub1[10.240.0.4] => alb[Potential LB private IP][10.240.1.0/25] : All Connections
```
ShiriMoran, line 79: Here the dst pip are not the same as before, same question

pkg/ibmvpc/analysis_output_test.go
```diff
@@ -704,10 +704,10 @@ var tests = []*vpcGeneralTest{
...
 	// 2. after abstraction, all connections from/to the LB are marked with **, (i.e. the abstraction did over approximation)
 	// for example here only four private IPs are connected to vsi1-sub3[10.240.128.5]:
-	//     vsi1-sub3[10.240.128.5] => alb[Fake LB private IP][10.240.1.128] : All Connections
-	//     vsi1-sub3[10.240.128.5] => alb[Fake LB private IP][10.240.1.4] : All Connections
-	//     vsi1-sub3[10.240.128.5] => alb[Fake LB private IP][10.240.129.128] : All Connections
-	//     vsi1-sub3[10.240.128.5] => alb[Fake LB private IP][10.240.65.128] : All Connections
+	//     vsi1-sub3[10.240.128.5] => alb[Potential LB private IP][10.240.1.128] : All Connections
+	//     vsi1-sub3[10.240.128.5] => alb[Potential LB private IP][10.240.1.4] : All Connections
+	//     vsi1-sub3[10.240.128.5] => alb[Potential LB private IP][10.240.129.128] : All Connections
+	//     vsi1-sub3[10.240.128.5] => alb[Potential LB private IP][10.240.65.128] : All Connections
```
adisos, line 710: update output lines with cidrs instead of address for potential private ip

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,360 @@
...
+		network ACL ky-test-private-2-others-acl allows connection with the following allow rules
+			index: 0, direction: inbound , src: 192.168.32.0/20 , dst: 0.0.0.0/0, conn: all, action: allow
+		security group kube-clusterid:1 allows connection with the following allow rules
+			index: 3, direction: inbound,  conns: protocol: tcp,  dstPorts: 30000-32767, remote: 0.0.0.0/0, local: 0.0.0.0/0
+			index: 4, direction: inbound,  conns: protocol: udp,  dstPorts: 30000-32767, remote: 0.0.0.0/0, local: 0.0.0.0/0
+			index: 6, direction: inbound,  conns: protocol: icmp,  icmpType: protocol: ICMP icmp-type: 8, remote: 0.0.0.0/0, local: 0.0.0.0/0
+		security group ky-test-default-sg allows connection with the following allow rules
+			index: 2, direction: inbound,  conns: protocol: all, remote: 0.0.0.0/0, local: 0.0.0.0/0
+
+TCP response is enabled; The relevant rules are:
```
ShiriMoran, line 34: The "LB rule" is indeed ill-relevant for TCP respond? @adisos @kyorav

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -344,7 +345,7 @@ func (c *VPCConfig) getNodesOfVsi(name string) ([]Node, int, error) {
 		vpc = cidrOrNameSlice[0]
 		vsi = cidrOrNameSlice[1]
 	}
-	for _, nodeSet := range c.NodeSets {
+	for _, nodeSet := range append(c.NodeSets, c.loadBalancersAsNodeSets()...) {
```
adisos, line 348: the function name is now misleading (`getNodesOfVsi`), it returns also LB nodes, not only VSI nodes

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,360 @@
+Explaining connectivity from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca to 192.168.4.4 (iks-node[192.168.4.4]) within ky-test-vpc
+========================================================================================================================================
```
ShiriMoran, line 2: The user queried the path between two single addresses and got 9 (!) different paths.
This is unique to the LB.
Perhaps the incentive here  should be explained? add the explanation right after the header?

ShiriMoran: Also, I did not verify the correctness of the output w.r.t. the input files.

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,296 @@
...
+
+No connections are allowed from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.36.6] to iks-clusterid:1[192.168.32.5];
+connection is blocked by Load Balancer
+
+Load Balancer: kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca blocks connection to iks-clusterid:1[192.168.32.5] which is not one of its pool members
+Egress: security group kube-r006-d7cfb31a-1d4b-40c8-83df-ce2e6f8f2e57 allows connection; network ACL ky-test-edge-acl allows connection
+Ingress: network ACL ky-test-edge-acl allows connection; security group kube-r006-d7cfb31a-1d4b-40c8-83df-ce2e6f8f2e57 allows connection
+
+Path:
+	kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.36.6] -> |
```
adisos, line 12: in this case maybe we do not need to split the output per private IP ?
the user asks about LB, and if the dest is not a pool member, it is a common explanation why the connection is blocked for all private IPs ?
or add initial such explanation for the LB itself, and then per private IP the explanation per sg/acl ?

adisos (commented):
Added few comments

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,296 @@
+Explaining connectivity from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca to 192.168.32.5 (iks-clusterid:1[192.168.32.5]) within ky-test-vpc
+=================================================================================================================================================
+
```
ShiriMoran, line 3: I'm not sure that in this specific case we need to split to all possible sub-subnets as per sg/nacl different rules combination. All paths are disabled since the target is not a pool member. Would the user care to see all the different rules combinations which are ill-relevant anyways?

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,296 @@
...
+No connections are allowed from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.36.6] to iks-clusterid:1[192.168.32.5];
+connection is blocked by Load Balancer
+
+Load Balancer: kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca blocks connection to iks-clusterid:1[192.168.32.5] which is not one of its pool members
```
ShiriMoran, line 7: Why is the LB not reflected in the name here?

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,360 @@
+Explaining connectivity from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca to 192.168.4.4 (iks-node[192.168.4.4]) within ky-test-vpc
```
ShiriMoran, line 1: Why is the LB not reflected in the name here?

pkg/ibmvpc/explainability_test.go
```diff
@@ -867,3 +883,17 @@ func TestMultiExplainabilityOutput(t *testing.T) {
...
+	vpcConfigMultiVpc := getConfig(t, "iks_config_object")
+	require.NotNil(t, vpcConfigMultiVpc, "vpcConfigMultiVpc equals nil")
+
+	pipCidr := "192.168.36.6"
+	cidr2 := "192.168.4.4"
+	// should fail since pip address can not be an explainability input
+	_, err1 := vpcConfigMultiVpc.ExplainConnectivity(pipCidr, cidr2, nil)
+	fmt.Println(err1.Error())
+	require.NotNil(t, err1, "the test should fail since "+pipCidr+" is a Private IP address")
+	require.Equal(t, "illegal src: no network interfaces are connected to "+pipCidr, err1.Error())
```
ShiriMoran, line 897: Is it suffice to provide the "usual" error message here? Would it help to mention this is a potential pip?

pkg/ibmvpc/parser.go
```diff
@@ -1655,7 +1655,7 @@ func getLoadBalancerIPs(vpcConfig *vpcmodel.VPCConfig,
 // also update vpcConfig & subnet with the result
 func createPrivateIP(name, id, address, publicAddress string,
 	vpc vpcmodel.VPCResourceIntf, loadBalancer *LoadBalancer, vpcConfig *vpcmodel.VPCConfig,
-	original bool, subnet *Subnet) (*PrivateIP, error) {
+	original bool, block *ipblock.IPBlock, subnet *Subnet) (*PrivateIP, error) {
```
ShiriMoran, line 1658: The privateIP per block is not new to this PR. Why this change now?

pkg/vpcmodel/explainabilityInput.go
```diff
@@ -395,6 +396,8 @@ func (c *VPCConfig) getNodesFromAddress(ipOrCidr string, inputIPBlock *ipblock.I
 	}
 	// internal address
 	networkInterfaces := c.GetNodesWithinInternalAddress(inputIPBlock)
+	// filtering out the nodes which are not represented by their address (currently only LB private IPs):
+	networkInterfaces = slices.DeleteFunc(networkInterfaces, func(n Node) bool { return !n.RepresentedByAddress() })
```
ShiriMoran, line 400: Should this specific case be treated differently in the relevant error message?

ShiriMoran (changes requested):
First (and main) batch of comments

pkg/vpcmodel/explainabilityInput.go
haim-kermany: yes, the function is not just misleading, it is also wrong:
1. it returns also vpe nodes
2. it is part of the vpcmodel, a package that should not know what a vsi, or vpe is

the task of separating the ibm specific references from vpcmodel package is out of the scope of this PR.

pkg/ibmvpc/vpc.go
```diff
@@ -70,18 +70,23 @@ type PrivateIP struct {
 	vpcmodel.VPCResource
 	vpcmodel.InternalNode
 	loadBalancer *LoadBalancer
-	// Since not all the LB balancer has a private IP, we create fake Private IPs at the subnets that do not have one.
-	// original - does the private IP was originally at the config file, or it is a fake one
+	// Since not all the LB balancer has a private IP, we create a potential Private Ip at the subnets that do not have one.
+	// original - does the private IP was originally at the config file, or is it a potential one
 	original bool
+	// the block which the pip was created for:
```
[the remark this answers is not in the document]

pkg/ibmvpc/analysis_output_test.go
haim-kermany: fixed

pkg/ibmvpc/examples/out/analysis_out/filters_split_lb_subnet_all_vpcs__no_lbAbstract.txt
haim-kermany: yes, it is intended. its not relevant to the cli.
it make more sense for fake/potenetial private IPs to print the block that they are created for, and not the specific address that was chosen for them

pkg/ibmvpc/examples/out/analysis_out/filters_split_lb_subnet_all_vpcs__no_lbAbstract.txt
haim-kermany: same answer

pkg/drawio/iconTreeNode.go
```diff
@@ -260,7 +260,7 @@ func newLoadBalancerTreeNode(parent SquareTreeNodeInterface, name string, privat
 
 type PrivateIPTreeNode struct {
 	abstractIconTreeNode
-	original bool // does the private IP was originally at the config file, or it is a fake one
+	original bool // does the private IP was originally at the config file, or is it just potential one
```
ShiriMoran, line 263: is it an potential address or the entire relevant cidr? in the reports it seems the latter

pkg/ibmvpc/vpc.go
```diff
@@ -70,18 +70,23 @@ type PrivateIP struct {
 	vpcmodel.VPCResource
 	vpcmodel.InternalNode
 	loadBalancer *LoadBalancer
-	// Since not all the LB balancer has a private IP, we create fake Private IPs at the subnets that do not have one.
-	// original - does the private IP was originally at the config file, or it is a fake one
+	// Since not all the LB has a private IP, we create a potential Private IP at the LB's subnets that do not have one.
```
ShiriMoran, line 73: Again, now you are presenting the entire relevant cidr, right?
This should be reflected in the documentation

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
haim-kermany: here it is the input, as it was got from the user.

pkg/ibmvpc/explainability_test.go
haim-kermany: also the addresses of non potential private ip are not allowed to be inputs

pkg/ibmvpc/parser.go
haim-kermany: since we add it to the private ip name

pkg/vpcmodel/explainabilityInput.go
ShiriMoran: The (original) function returns Nodes representing an endpoint given by name. It is not ibm specific and has the correct abstraction level. Indeed, the names (of the func and of the variables) are misleading. The function is not wrong and no need for a PR in the regard.

pkg/drawio/iconTreeNode.go
haim-kermany: there is no change in the implementation, when creating a fake IP, we allocate for it one address, we just made an adjustment in the report - instead of printing the allocated address, we print the block

pkg/ibmvpc/vpc.go
haim-kermany: there is no change in the implementation, when creating a fake IP, we allocate for it one address, we just made an adjustment in the report - instead of printing the allocated address, we print the block

pkg/vpcmodel/explainabilityInput.go
haim-kermany: there isnt really a specific case here, it just that GetNodesWithinInternalAddress() might give us some nodes that we do not want to

pkg/vpcmodel/explainabilityInput.go
haim-kermany: current code and documentation *was* not consistent and misleading.
1. If I will change the function name, to context of the function call will be misleading
2. all what you ask here is out of the scope of this PR. 
3. I added a todo comment, you are welcome to follow it

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
haim-kermany: the load balancer is is being treat as all other nodeSets, I didnt find an explanability example, so I added multiNIsToSingleNI.
I also played with vpe "iks-clusterid:1" at iks_config_object example
in such cases, the explanation is being split to the different NIs, even when the explanation is the same for all NIs/ResIp
there is a need to group explanations, but since we want to show all the relevant rules, the solution can not be load balancer specific

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
haim-kermany: same answer

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
haim-kermany: in case a nodeSet is an input, we only write the user input. also see:
Explaining connectivity from vsi3-ky to vsi1-ky within test-vpc1-ky
I can add the kind here, but the solution will apply for all kind of nodeSets. @adisos ?

pkg/vpcmodel/explainabilityInput.go
adisos: @ShiriMoran , will open an issue for you to revisit this function.

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
adisos: please open a separate issue so this can be considered there.

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
adisos: I think adding the kind of nodeSets can improve the report.

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
haim-kermany: https://github.com/np-guard/vpc-network-config-analyzer/issues/674

pkg/vpcmodel/explainabilityInput.go
ShiriMoran: The function is consistent; some documentation and renaming could indeed help.

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
haim-kermany: all other nodeset can be suffer from the case of "one rule control them all", for example, two nodesets of the size of 4 which are not on the same vpc - you will get the exact same explain 16 times.
I dont think it is good idea to make a special implementation fo load balancer

adisos: @kyorav @zivnevo what do you think? please review the explainability output for LB

adisos: > I'm not sure that in this specific case we need to split to all possible sub-subnets as per sg/nacl different rules combination. All paths are disabled since the target is not a pool member. Would the user care to see all the different rules combinations which are ill-relevant anyways?

I'd not consider sg/nacl rules as necessarily irrelevant in this case, since it might be relevant to check if the connection is blocked only by LB configuration or also by nacl/sg configs.

pkg/vpcmodel/abstractVPC.go
```diff
@@ -113,6 +113,8 @@ type Node interface {
 	// AbstractedToNodeSet returns the abstracted nodeSet that contains this node (if any)
 	// e.g. the Nodes of Load Balancer private IPs are abstracted by the Load Balancer
 	AbstractedToNodeSet() NodeSet
+	// RepresentedByAddress - does the node be identify from a query address
```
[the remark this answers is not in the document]
haim-kermany: fixed

adisos (approved):

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,360 @@
...
+Path:
+	kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.36.6] -> security group kube-r006-d7cfb31a-1d4b-40c8-83df-ce2e6f8f2e57 -> ky-test-edge-subnet-2 -> network ACL ky-test-edge-acl -> 
+	network ACL ky-test-private-2-others-acl -> ky-test-private-subnet-2 -> security group[kube-clusterid:1, ky-test-default-sg] -> iks-node[192.168.4.4]
+
+
+Details:
+~~~~~~~~
+Path is enabled; The relevant rules are:
+	Load Balancer:
+		kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LoadBalancer] allows connection to iks-node[192.168.4.4] which is one of its pool members
```
kyorav, line 15:
```suggestion
		kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LoadBalancer] may initiate a connection to iks-node[192.168.4.4], which is one of its pool members
```

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,360 @@
...
+		network ACL ky-test-private-2-others-acl allows connection with the following allow rules
+			index: 4, direction: outbound , src: 0.0.0.0/0 , dst: 192.168.32.0/20, conn: all, action: allow
+
+	Ingress:
+		network ACL ky-test-edge-acl allows connection with the following allow rules
+			index: 0, direction: inbound , src: 0.0.0.0/0 , dst: 0.0.0.0/0, conn: all, action: allow
+
+------------------------------------------------------------------------------------------------------------------------
+
+Allowed connections from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[Potential LB private IP][192.168.0.0/22] to iks-node[192.168.4.4]: protocol: TCP,UDP dst-ports: 30000-32767
```
kyorav, line 86: What is the logic for the creation of a potential LB private IP? If this is an endpoint that we invented I'd expect to see a nicer and shorter name for it, maybe concatenate the LB name with `_ip_` and the private IP index?

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,296 @@
+Explaining connectivity from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca to 192.168.32.5 (iks-clusterid:1[192.168.32.5]) within ky-test-vpc
+=================================================================================================================================================
+
+No connections are allowed from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.36.6] to iks-clusterid:1[192.168.32.5];
+connection is blocked by Load Balancer
+
+Load Balancer: kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LoadBalancer] blocks connection to iks-clusterid:1[192.168.32.5] which is not one of its pool members
```
kyorav, line 7: In the case where the source is a LB and the destination is not a pool member I would print a message along the lines of 
```
The load balancer will not connect to this endpoint since it is not a pool member
```
and nothing else (drop the explanation about SGs and nACLs). I doubt users are interested in an analysis of the nACLs and SGs in this case. 

Either way, if you choose to elaborate SGs and nACLs on the way, do not use the phrasing "blocks connection" because the LB is not a firewall and cannot actually block connectivity.

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
ShiriMoran: #674

pkg/ibmvpc/examples/out/explain_out/multiNIsToSingleNI_all_vpcs_explain_debug.txt
```diff
@@ -0,0 +1,58 @@
+Explaining connectivity from vsi3-ky to vsi1-ky within test-vpc1-ky
```
ShiriMoran, line 1: What is this test about?

ShiriMoran (dismissed):

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
haim-kermany: its <lb name>[<end point type>][<ip range>]. the ip range is the rang that we create the pip for.
yes, in this example the name is not pretty.

pkg/ibmvpc/examples/out/explain_out/LBToIksNode_all_vpcs_explain_debug.txt
haim-kermany: fixed

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
```diff
@@ -4,7 +4,7 @@ Explaining connectivity from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca t
 No connections are allowed from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.36.6] to iks-clusterid:1[192.168.32.5];
 connection is blocked by Load Balancer
```
adisos, line 5: this phrasing is also an issue for LB , connection is not "blocked" by load balancer , and the term "no connections are allowed" is also an issue for LB context

adisos: I would replace "no connections are allowed from .." with just "no connections from ..." 

and remove the sentence "connection is blocked by Load Balancer"

adisos: for the other LB example  - would replace "Allowed connections from .. " to "connections from ... "

pkg/ibmvpc/examples/out/analysis_out/actual.txt
adisos, line 1: remove this file

pkg/ibmvpc/examples/out/analysis_out/expected.txt
adisos, line 1: remove this file

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
```diff
@@ -35,7 +35,7 @@ Path is disabled; The relevant rules are:
 ------------------------------------------------------------------------------------------------------------------------
 
 No connections from from kube-clusterid:1-8fdd1d0a2ce34deba99d0f885451b1ca[LB private IP][192.168.40.6] to iks-clusterid:1[192.168.32.5];
-connection is blocked by Load Balancer
+connection can not be initiated by Load Balancer
```
adisos, line 38: "can not be initiated " => "will not be initiated"

pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -204,37 +204,37 @@ func (g *groupedConnLine) explainPerCaseStr(c *VPCConfig, src, dst EndpointElem,
...
 	default: // there is a connection
 		return existingConnectionStr(c, connQuery, src, dst, conn, path, details)
 	}
 }
 
-// blockedByString() return the block string according to the flags
-func blockedByString(ingressBlocking, egressBlocking, loadBalancerBlocking bool) string {
+// blockSummery() return a summery of the rules that block the connection, for example:
+// "connection is blocked both by ingress and egress, and can not be initiated by Load Balancer"
+func blockSummery(ingressBlocking, egressBlocking, loadBalancerBlocking bool) string {
```
adisos, line 217:
```suggestion
// blockSummary() return a summary of the rules that block the connection, for example:
// "connection is blocked both by ingress and egress, and cannot be initiated by Load Balancer"
func blockSummary(ingressBlocking, egressBlocking, loadBalancerBlocking bool) string {
```

try checking typo checks plugin for vscode

pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -262,7 +262,7 @@ func crossVpcRouterRequired(src, dst EndpointElem) bool {
 
 // returns string of header in case a connection fails to exist
 func noConnectionHeader(src, dst string, connQuery *connection.Set) string {
-	return fmt.Sprintf("No connections are allowed from %s to %s%s;", src, dst, connHeader(connQuery))
+	return fmt.Sprintf("No connections from from %s to %s%s;", src, dst, connHeader(connQuery))
```
adisos, line 265:
```suggestion
	return fmt.Sprintf("No connections from %s to %s%s;", src, dst, connHeader(connQuery))
```

adisos (changes requested):

pkg/ibmvpc/examples/out/explain_out/GroupingExternalSG1_all_vpcs_explain.txt
```diff
@@ -10,7 +10,7 @@ Path:
 
 ------------------------------------------------------------------------------------------------------------------------
 
-No connections are allowed from vsi1-ky[10.240.10.4] to Public Internet 161.0.0.0-161.25.255.255,161.27.0.0-161.255.255.255;
+No connections from from vsi1-ky[10.240.10.4] to Public Internet 161.0.0.0-161.25.255.255,161.27.0.0-161.255.255.255;
```
ShiriMoran, line 13: one "from" will suffice here

pkg/ibmvpc/examples/out/explain_out/GroupingExternalSG1_all_vpcs_explain.txt
```diff
@@ -1,7 +1,7 @@
 Explaining connectivity from vsi1-ky to 161.26.0.0/8 within test-vpc1-ky
 ========================================================================
 
-Allowed connections from vsi1-ky[10.240.10.4] to Public Internet 161.26.0.0/16: protocol: UDP
+Connections from vsi1-ky[10.240.10.4] to Public Internet 161.26.0.0/16: protocol: UDP
```
ShiriMoran, line 4: Does the new sentence make sense? That allows connections from src to dst are UDP? 
I think not (it was clear before)

pkg/ibmvpc/examples/out/explain_out/LBToResIPNode_all_vpcs_explain_debug.txt
haim-kermany: fixed

pkg/vpcmodel/explainabilityPrint.go
haim-kermany: fixed

pkg/vpcmodel/explainabilityPrint.go
haim-kermany: fixed

pkg/ibmvpc/examples/out/explain_out/GroupingExternalSG1_all_vpcs_explain.txt
haim-kermany: fixed

pkg/ibmvpc/examples/out/explain_out/GroupingExternalSG1_all_vpcs_explain.txt
adisos: I think it is clear enough.. the term "allows" is not general enough, for LB it's not a matter of "allowing"

adisos (approved):

PR #675, opened 2024-07-07
116 lint intrf+first

ShiriMoran:


pkg/linter/linterExecute.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
+)
+
+const issues = " issues:"
+
+// LinterExecute executes linters one by one
+// todo: mechanism for disabling/enabling lint checks
+func LinterExecute(configsMap map[string]*vpcmodel.VPCConfig) (issueFound bool, resString string) {
+	strPerVpc := []string{}
+	for _, config := range configsMap {
```
zivnevo, line 23: There is a design decision here: linters can only look at one config at a time.
Might there be a linter that needs to consider multiple configs together?

pkg/linter/lintFilterRuleSplitSubnet.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, ruleIPBlock := range ruleIPBlocks {
+		if ruleIPBlock.Overlap(subnetCidrIPBlock) && !subnetCidrIPBlock.ContainedIn(ruleIPBlock) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (lint *filterRuleSplitSubnet) getName() string {
+	return "Firewalls' rules that splits subnets"
```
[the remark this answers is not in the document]
ShiriMoran: This is the **name** of the lint. Shouldn't be this long. Perhaps it makes sense to add a description to the interface?

pkg/linter/linterExecute.go
ShiriMoran: 1. I don't have one in mind
2. I would expect lint to be relatively local
3. And yet, better have a flexible framework. I will change the interface as you suggest.

pkg/linter/linterTypes.go
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+package linter
+
+import "github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
+
+type linter interface {
+	check() ([]string, error) // returns []string of size 0 if no lint issues
+	getName() string
+	getFindings() []any
```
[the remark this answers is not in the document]
ShiriMoran: sounds right. Just one comment: this implies a very local context in each lint msg. Namely, each lint msg will contain the vpc name in addition to the other details - instead of having a list of all linting comments of each vpc

pkg/linter/lintFilterRuleSplitSubnet.go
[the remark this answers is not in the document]
zivnevo: Right. Let's have both `Name()` and `Description()`.
Name can then be really short. Something like `rules-splitting-subnets`

ShiriMoran:
1. Printing can be improved - e.g. print only once VPC name for all relevant issues. Not in this PR
2. toJSON still not implemented

pkg/linter/linterExecute.go
ShiriMoran: done

pkg/linter/linterTypes.go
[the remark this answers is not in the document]
ShiriMoran: done

pkg/linter/linterTypes.go
```diff
@@ -0,0 +1,28 @@
...
+Copyright 2023- IBM Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package linter
+
+import "github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
+
+// todo: export certain functionality
```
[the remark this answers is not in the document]
ShiriMoran: Agreed. comment removed and function de-imported

pkg/linter/linterTypes.go
```diff
@@ -0,0 +1,28 @@
...
+type linter interface {
+	Check() ([]finding, error) // false if issues found
+	LintName() string          // this lint Name
+	LintDescription() string   // this string Name
+	String() string            // string with this lint's finding
+	ToJSON() []any             // this lint finding in JSON
```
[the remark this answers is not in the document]
ShiriMoran: Not sure I follow. Why is `Findings() []finding` needed in this interface?  and what does it has to do with the aggregation? The aggregation is in an upper level? perhaps we need another interface here? in the aggregation level?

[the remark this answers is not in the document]
ShiriMoran: added function getFindings() (there is a member named finding() can't a have a function with the same name)

pkg/linter/linterTypes.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type finding interface {
+	vpc() string
+	string() string
+	toJSON() any
+}
+
+type basicLinter struct {
+	configs map[string]*vpcmodel.VPCConfig
```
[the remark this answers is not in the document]
ShiriMoran: added `findings []finding` member and a standard `addFinding(f finding)` method

pkg/ibmvpc/examples/out/lint_out/basic_sg1_Lint
```diff
@@ -0,0 +1,8 @@
+"Firewall rules implying different connectivity for different endpoints within a subnet" issues:
+------------------------------------------------------------------------------------------------
+In VPC test-vpc1-ky, Network acl sg1-ky rule's indexed 1 splits subnet subnet1-ky (10.240.10.0/24). Splitting rule details: index: 1, direction: inbound,  conns: protocol: all, remote: sg1-ky (10.240.10.4/32), local: 0.0.0.0/0
```
adisos, line 3: `Network acl sg1-ky` should be replaced with `security group sg1-ky` ?

pkg/vpcmodel/abstractVPC.go
```diff
@@ -246,6 +246,20 @@ type RulesInTable struct {
 	RulesOfType RulesType
 }
 
+// RuleOfFilter a single rule in filter given the layer (SGLayer/NACLLayer)
+type RuleOfFilter struct {
+	LayerName  string             `json:"layer"`
+	FilterName string             `json:"table"`
+	RuleIndx   int                `json:"rule_index"`
+	RuleDesc   string             `json:"rule_description"`
+	IPBlocks   []*ipblock.IPBlock `json:"ip_blocks,omitempty"`
```
adisos, line 255: what is the `IPBlocks` field? how is it related to the rule? missing some description

pkg/vpcmodel/abstractVPC.go
```diff
@@ -246,6 +246,20 @@ type RulesInTable struct {
 	RulesOfType RulesType
 }
 
+// RuleOfFilter a single rule in filter given the layer (SGLayer/NACLLayer)
+type RuleOfFilter struct {
+	LayerName  string             `json:"layer"`
+	FilterName string             `json:"table"`
+	RuleIndx   int                `json:"rule_index"`
```
adisos, line 253: typo `RuleIndx` => `RuleIndex`

adisos (commented):
Added few comments

pkg/ibmvpc/examples/out/lint_out/basic_sg1_Lint
ShiriMoran: obviously.... fixed

pkg/vpcmodel/abstractVPC.go
ShiriMoran: The IPBlocks are the cidrs/IPBlocks the Rule refers to (src/dst/local/... however it is called in each filter).
Documentation added

pkg/vpcmodel/abstractVPC.go
ShiriMoran: it was on purpose, not a typo. No big deal  - changed to `RuleIndex`.

pkg/linter/lintFilterRuleSplitSubnet.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil
+}
+
+// given a subnet and IPBlocks mentioned in a rule, returns the list
+func ruleSplitSubnet(subnet vpcmodel.Subnet, ruleIPBlocks []*ipblock.IPBlock) (bool, error) {
+	cidr := subnet.CIDR()
+	subnetCidrIPBlock, err := ipblock.FromCidr(cidr)
+	if err != nil {
+		return false, err
+	}
```
adisos, line 80:
```suggestion
	subnetCidrIPBlock := subnet.AddressRange()
```

pkg/linter/lintFilterRuleSplitSubnet.go
```diff
@@ -0,0 +1,135 @@
...
+		if config.IsMultipleVPCsConfig {
+			continue // no use in executing lint on dummy vpcs
+		}
+		for _, layer := range vpcmodel.FilterLayers {
+			filterLayer := config.GetFilterTrafficResourceOfKind(layer)
+			rules, err := filterLayer.GetRules()
+			if err != nil {
+				return err
+			}
+			for _, rule := range rules {
```
adisos, line 53: if a rule in acl is splitting a subnet, but it has no effect on connectivity (for example if that nacl is not attached to the relevant subnet), it is also going to be reported?

ShiriMoran: Yes, it is reported.
The question if it report always as now or to impose more filters is not trivial. Whether we report on potential problems or real problems.
1. I would surely not take consideration as redundant rule. We will have a different lint for Redundant rules.
2. `NACL` wise - should we consider only `relevant subnets` (which of course has to be defined)? Similar question can also be asked about SGs.  

I tend to think that we should keep it as it is now, for a few reasons:
1. We report best practices; it is up to the user to decide if its a problem or not. And breaking a subnet does not seem best practice.
1. Keep it simple and easy to define and understand. 

At the end of the day we should have many lints, each can be enabled or disabled. So perhaps we can have another, stricter, lint that reports only rules that results in different connectivity in the same subnet.

pkg/linter/lintFilterRuleSplitSubnet.go
ShiriMoran: right. done.

pkg/linter/lintFilterRuleSplitSubnet.go
kyorav: I tend to agree with @ShiriMoran that linting should be simplistic and represent best practice even if it has no effect. Lots of examples like that in code linters.

adisos (approved):

PR #688, opened 2024-07-09
basic support for synthesis output format

olasaadi99:


kyorav (commented):
A few things:
- The output needs to go to the file specified by the `--output-file` flag
- Names should be as they appear in the configuration object. For example, instead of `vsi0-ctrl-sub2[10.240.66.4]` you should use `vsi0-ctrl-sub2`
- The `type` field should be according to the spec language JSON schema, see [here](https://github.com/np-guard/models/blob/668e216f9f9536e06f606997e5a656e029416c70/spec_schema.json#L119)
- External CIDRs need to be defined as named externals. See an example of how this looks like [here](https://github.com/np-guard/vpc-network-config-synthesis/blob/main/test/data/sg_externals/conn_spec.json). If there are multiple different external CIDRs they must have different names.

kyorav (changes requested):
The default grouping of externals into a single "public internet" needs to be applied for the synthesis output, creating a single external with the cidr "0.0.0.0/0".

See for example [this example](https://github.ibm.com/np-guard/terraform-examples/tree/main/acl_testing4)

pkg/commonvpc/vpc.go
```diff
@@ -46,7 +46,16 @@ func zoneFromVPCResource(r vpcmodel.VPCResourceIntf) (*Zone, error) {
 type NetworkInterface struct {
 	vpcmodel.VPCResource
 	vpcmodel.InternalNode
-	Vsi string `json:"-"`
+	Vsi               string `json:"-"`
+	numberOfNifsInVsi int
```
zivnevo, line 50: How about storing just a pointer to the parent `commonvpc.Vsi` instead of these two fields?
You can then use `parentVSI.Name` and `len(parentVsi.VPCnodes)` to obtain these two values.

zivnevo (changes requested):

pkg/commonvpc/vpc.go
olasaadi99: a lot of tests use vsi name but does not init the vsi struct, that need a big change

olasaadi99:
test is not stable, will be handled next commit

kyorav (commented):
I noticed the output contains duplicate required connections. For example, if you run it on (the latest) `acl_testing4` you see things like this: 
```
{
            "allowed-protocols": [
                {
                    "protocol": "ANY"
                }
            ],
            "dst": {
                "name": "vsi3a",
                "type": "instance"
            },
            "src": {
                "name": "vsi3b",
                "type": "instance"
            }
        },
        {
            "dst": {
                "name": "vsi3a",
                "type": "instance"
            },
            "src": {
                "name": "vsi3b",
                "type": "instance"
            }
        }
```

I believe the synthesis logic is smart enough not to generate superfluous rules, but it's probably better not to generate them in the first place.  @zivnevo what do you think?

olasaadi99:
> I noticed the output contains duplicate required connections. For example, if you run it on (the latest) `acl_testing4` you see things like this:
> 
> ```
> {
>             "allowed-protocols": [
>                 {
>                     "protocol": "ANY"
>                 }
>             ],
>             "dst": {
>                 "name": "vsi3a",
>                 "type": "instance"
>             },
>             "src": {
>                 "name": "vsi3b",
>                 "type": "instance"
>             }
>         },
>         {
>             "dst": {
>                 "name": "vsi3a",
>                 "type": "instance"
>             },
>             "src": {
>                 "name": "vsi3b",
>                 "type": "instance"
>             }
>         }
> ```
> 
> I believe the synthesis logic is smart enough not to generate superfluous rules, but it's probably better not to generate them in the first place. @zivnevo what do you think?

fixed

kyorav:
The output seems good and the synthesis works on it. I will run a test to verify correctness.

kyorav:
I ran the analyzer on the attached object twice, once with `txt` output and once with `synthesis`, and they don't match. The `txt` output contains the following connection (that I think is correct) and does not show up as a required connection
```
proxy[10.240.0.4] => appdata-endpoint-gateway[10.240.0.5] : protocol: TCP dst-ports: 9000-9999
```

[origin_config.json](https://github.com/user-attachments/files/16609608/origin_config.json)

olasaadi99:
> I ran the analyzer on the attached object twice, once with `txt` output and once with `synthesis`, and they don't match. The `txt` output contains the following two connections (that I think are correct) and do not show up as a required connection
> 
> ```
> be[10.240.128.6] => policydb-endpoint-gateway[10.240.128.8] : protocol: TCP
> proxy[10.240.0.4] => appdata-endpoint-gateway[10.240.0.5] : protocol: TCP dst-ports: 9000-9999
> ```
> 
> [origin_config.json](https://github.com/user-attachments/files/16609608/origin_config.json)

if you look into warnings you can see the following
`2024/08/14 12:03:25 WARN        skipping resource appdata-vpe1, err: resourceType ReservedIP is not supported yet
2024/08/14 12:03:26 WARN        skipping resource policydb-vpe2, err: resourceType ReservedIP is not supported yet`
we agreed that reserved ips are not supported yet right?

zivnevo:
> > I ran the analyzer on the attached object twice, once with `txt` output and once with `synthesis`, and they don't match. The `txt` output contains the following two connections (that I think are correct) and do not show up as a required connection
> > ```
> > be[10.240.128.6] => policydb-endpoint-gateway[10.240.128.8] : protocol: TCP
> > proxy[10.240.0.4] => appdata-endpoint-gateway[10.240.0.5] : protocol: TCP dst-ports: 9000-9999
> > ```
> >   
> > [origin_config.json](https://github.com/user-attachments/files/16609608/origin_config.json)
> 
> if you look into warnings you can see the following `2024/08/14 12:03:25 WARN skipping resource appdata-vpe1, err: resourceType ReservedIP is not supported yet 2024/08/14 12:03:26 WARN skipping resource policydb-vpe2, err: resourceType ReservedIP is not supported yet` we agreed that reserved ips are not supported yet right?

I think that everything that is supported on the text connectivity report, should be supported here too.

pkg/vpcmodel/resourceTypes.go
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright 2023- IBM Inc. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package vpcmodel
+
+// Resource types const strings, used in the generated resources of this pkg
+const (
```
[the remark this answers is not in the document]
olasaadi99: did this to convert types to syntheses format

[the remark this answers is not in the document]
olasaadi99: is that ok or should I move the handleTypes function to commonvpc (but this pkg is not relevant to synthesis)

kyorav (commented):
Sanity checked passed! I was able to produce synthesis output, run it through generation, deploy the generated SGs, and get an equivalent network connectivity report.

zivnevo:
Some decisions from today's discussion:
* Replace the `DrawioResourceIntf` with a new interface, called `FormattableResource`.
* Put the new interface in a new Go file `formattableResource.go`
* Add to the new interface two syntehsis-output-specific functions: `DetailedResourceForSynthesisOut()` and `SynthesisKind()`
* See if this allows us to remove `resourceTypes.go` from `vpcmodel`
* We need to reconsider interface hierarchy (opened issue #801)

pkg/vpcmodel/formattableResource.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package vpcmodel
+
+import (
+	"github.com/np-guard/models/pkg/spec"
+	"github.com/np-guard/vpc-network-config-analyzer/pkg/drawio"
+)
+
+// FormattableResource is the interface of all the resources that are converted to a drawio treeNodes
```
zivnevo, line 14:
```suggestion
// FormattableResource is an interface for resources that appear in the tool's output, and for which the printed text may change according to the selected format
```

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// if this nif's vsi has only one nif, we convert it to instance type with name of the instance
+	// because the name of the nif will be meaningless for the user if there is one generated nif.
+	if resourceType == spec.ResourceTypeNif && nifNumber == 1 {
+		resourceType = spec.ResourceTypeInstance
+	}
+	return
+}
+
+func GetSpec(groupedLines []*groupedConnLine) spec.Spec {
```
zivnevo, line 73:
```suggestion
func GetSynthesisSpec(groupedLines []*groupedConnLine) spec.Spec {
```

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -0,0 +1,153 @@
...
+	externalsMap := make(map[string]string)
+	sortGroupedLines(groupedLines)
+
+	for _, groupedLine := range groupedLines {
+		srcName, srcType := handleNameAndType(groupedLine.Src, externalsMap, externals)
+		dstName, dstType := handleNameAndType(groupedLine.Dst, externalsMap, externals)
+		if groupedLine.CommonProperties.Conn.isEmpty() {
+			continue
+		}
+		connLines = append(connLines, spec.SpecRequiredConnectionsElem{
```
zivnevo, line 86: Add a comment: "For now, ignoring responsiveness of TCP connection"

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -0,0 +1,153 @@
...
+			Src:              spec.Resource{Name: srcName, Type: srcType},
+			Dst:              spec.Resource{Name: dstName, Type: dstType},
+			AllowedProtocols: sortProtocolList(spec.ProtocolList(connection.ToJSON(groupedLine.CommonProperties.Conn.allConn)))})
+	}
+	s.Externals = externals
+	s.RequiredConnections = connLines
+	return s
+}
+
+func sortProtocolList(g spec.ProtocolList) spec.ProtocolList {
```
zivnevo, line 96: Any reason to do it so precisely?
How about a simpler sorting scheme: marshal each protocol into string, and sort the strings.
```golang
	sort.Slice(g, func(i, j int) bool {
		iMarshal, _ := json.Marshal(g[i])
		jMarshal, _ := json.Marshal(g[j])
		return string(iMarshal) > string(jMarshal)
	})

```

zivnevo (changes requested):

pkg/vpcmodel/synthesisOutput.go
olasaadi99: no reason, will change it

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -0,0 +1,116 @@
...
+			resourceName = handleExternals(resourceName, structObj.CidrOrAddress(), externalsMap, externals)
+		}
+	}
+	resourceType = resource.SynthesisKind()
+
+	// if this nif's vsi has only one nif, we convert it to instance type with name of the instance
+	// because the name of the nif will be meaningless for the user if there is one generated nif.
+	if resourceType == spec.ResourceTypeNif && nifNumber == 1 {
+		resourceType = spec.ResourceTypeInstance
+	}
```
zivnevo, line 70: Better move this code to `NetworkInterface::SynthesisKind()`

zivnevo (changes requested):

pkg/vpcmodel/output.go
```diff
@@ -293,6 +296,10 @@ func (of *serialOutputFormatter) AggregateVPCsOutput(outputList []*SingleAnalysi
 			all[o.VPC1Name] = o.jsonStruct
 		}
 		res, err = writeJSON(all, outFile)
+	case Synthesis:
+		// in synthesis format we need to follow json spec schema
+		// https://github.com/np-guard/models/blob/main/spec_schema.json
```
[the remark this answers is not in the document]
olasaadi99: @kyorav is multi-vpc supported?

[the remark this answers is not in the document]
zivnevo: Support for multi-vpc configs to be implemented in a future PR.
@olasaadi99 , please open an issue

zivnevo (approved):

kyorav (approved):

PR #691, opened 2024-07-09
690 explain path blocked

ShiriMoran:


ShiriMoran:
There is an issue with LoadBalancer:
It is not part of the path at the moment. Thus, when it is the blocking element it is not drawn as part of the path either.
Should we add it to the path ?? 
We should decide and then finalize the LoadBalancer tests (which are not in final state now)
@zivnevo @haim-kermany @adisos

zivnevo (approved):

adisos (approved):

PR #694, opened 2024-07-10
update cli flags + readme

adisos:
Changed flags:
```
-c, --config
-f, --filename
-o, --output
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Run the `vpcanalyzer` CLI tool with one of the following commands.
 ```
 
 ### Providing VPC configuration
-VPC configuration must be provided, using the `--config` option, as a `JSON` file produced by the [`cloud-resource-collector`](https://github.com/np-guard/cloud-resource-collector). Alternatively, VPC configuration can be read directly from a given account using the `--provider` flag.
+A VPC configuration must be provided, in one of two ways. Either the `--provider` flag is used, in which case the VPC configuration will be extracted directly from a given account, or a configuration object is provided by the user using the `--config` option. A configuration object can be independently produced by the [`cloud-resource-collector`](https://github.com/np-guard/cloud-resource-collector).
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -12,24 +12,25 @@ Run the `vpcanalyzer` CLI tool with one of the following commands.
...
+  -r, --region stringArray      cloud region from which to collect resources, can pass multiple regions
+      --resource-group string   resource group id or name from which to collect resources
+  -v, --verbose                 runs with more informative messages printed to log
+      --version                 version for vpcanalyzer
+      --vpc string              CRN of the VPC to analyze
 ```
 
 ### Providing VPC configuration
-VPC configuration should be provided, using the `--vpc-config` option, as a `JSON` file produced by the [`cloud-resource-collector`](https://github.com/np-guard/cloud-resource-collector). Alternatively, VPC configuration can be read directly from a given account using the `--provider` flag.
+VPC configuration must be provided, using the `--config` option, as a `JSON` file produced by the [`cloud-resource-collector`](https://github.com/np-guard/cloud-resource-collector). Alternatively, VPC configuration can be read directly from a given account using the `--provider` flag.
```
kyorav, line 30: This is confusing to me -- either you *must* provide a configuration with `--config`, or there is an alternative, which is it?
Maybe you mean "A VPC configuration must be provided, in one of two ways. Either the `--provider` flag is used, in which case the VPC configuration will be extracted directly from a given account, or a configuration object is provided by the user using the `--config` option. A configuration object can be independently produced by the [`cloud-resource-collector`](https://github.com/np-guard/cloud-resource-collector)"

zivnevo (dismissed):

README.md
adisos: right, updated.

kyorav (approved):

Issue #116, opened 2023-07-27
Implement Basic Linting


kyorav:
This issue is for brainstorming ideas for checks we may include as part of a linting functionality. At some point we will review the list with a stakeholder and decide which ones make sense.

1. partial connections. When a request can be sent to destination but the response will be blocked. #706 solves
1. nACL splits a subnet. When a cidr that does not cover the entire subnet is used in the source of an outbound rule or the destination of an inbound rule. #698 solves
2. sg split subnet as above. #698 solves
1. LB analysis: warn about SG that split subnet of a configured LB (without running full connectivity analysis)
4. As the above two items, but when there is an actual connectivity gap between two endpoints in the same subnet [define: what is considered an endpoint]
1. overlapping subnet ranges (relevant mostly for the multiple VPCs use case) #700 solves
1. source-port limitation may block response. This will help identify misuse of source-port. Note that this is relevant when the connection is complete. (@kyorav please elaborate) 
1. Non-connected tables: #722 solves
    1. SG not connected to an endpoint
    2. NACL not connected to a subnet
1. Non-connected Routers
    1. tgw not connected
    2. fip not connnected
    3. pgw not connected
    4. loadbalancer without pool members
1. Redundant Subnets (no connected endpoints)
2. Redundant VPC....
1. Redundant rules:
    1. SG/NACL rule that is useless w.r.t. the relevant address prefixes #115 solved by #731
    1. SG rule that is implied by other SG rules in the same table
    1. NACL rule that is implied by other, higher priories, NACL rules in the same table
    1. SG/NACL rules that do not actually effect the connectivity    
1. Routing Tables analysis: warn about non deterministic routing 
1. NACL rules with values in src/dst min/max port but no value in protocol - in which case the ports are ignored (discuss)

adisos:
Related: issue#115

ShiriMoran:
> This issue is for brainstorming ideas for checks we may include as part of a linting functionality. At some point we will review the list with a stakeholder and decide which ones make sense.
> 
>     * partial connections. When a request can be sent to destination but the response will be blocked.
> 
>     * nACL splits a subnet. When a cidr that does not cover the entire subnet is used in the source of an outbound rule or the destination of an inbound rule.
> 
>     * overlapping subnet ranges (relevant mostly for the multiple VPCs use case)
> 
>     * source-port limitation may block response. This will help identify misuse of source-port. Note that this is relevant when the connection is complete.
> 
[...]

PR #700, opened 2024-07-17
504 lint overlap subnets

ShiriMoran:
addresses #116


pkg/ibmvpc/examples/out/lint_out/acl3_overlap_subnets_Lint
```diff
@@ -0,0 +1,22 @@
...
+In VPC test-vpc1-ky, network ACL acl3-ky rule's indexed 1 splits subnets subnet1-ky (10.240.10.0/16), subnet3-ky (10.240.30.0/24). Splitting rule details: index: 1, direction: outbound , src: 10.240.30.0/31 , dst: 10.240.20.0/24, conn: all, action: allow
+In VPC test-vpc1-ky, network ACL acl3-ky rule's indexed 2 splits subnet subnet1-ky (10.240.10.0/16). Splitting rule details: index: 2, direction: inbound , src: 10.240.10.0/24 , dst: 0.0.0.0/0, conn: all, action: allow
+In VPC test-vpc1-ky, network ACL acl3-ky rule's indexed 3 splits subnets subnet1-ky (10.240.10.0/16), subnet3-ky (10.240.30.0/24). Splitting rule details: index: 3, direction: inbound , src: 10.240.20.0/24 , dst: 10.240.30.0/31, conn: all, action: allow
+________________________________________________________________________________________________________________________________________________________________________________________________________
+
+"Overlapping CIDR ranges between different subnets" issues:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+subnet subnet1-ky of cidr 10.240.10.0/16 and subnet subnet2-ky of cidr 10.240.20.0/24, both from VPC test-vpc1-ky, overlap in 10.240.20.0/24
```
adisos, line 21: isn't the goal to warn about subnets from different VPCs with overlapping ranges? 
I don't think it's possible to configure such subnets in the same VPC.

ShiriMoran: Yes, you are correct. It should not be possible. Redundant code and test removed.

adisos: please remove also the output file + add another relevant test instead

ShiriMoran: right. output file removed. Added a test in which the overlapping is in part of the range (in which case the output is slightly different)

pkg/linter/lintOverlappingSubnets.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (finding *overlapSubnets) string() string {
+	subnet1 := finding.overlapSubnets[0]
+	subnet2 := finding.overlapSubnets[1]
+	overlapStr := ""
+	if finding.overlapIPBlocks.String() == subnet1.CIDR() && subnet1.CIDR() == subnet2.CIDR() {
+		overlapStr = " overlap in the entire subnets' CIDR range"
+	} else {
+		overlapStr = fmt.Sprintf(" overlap in %s", finding.overlapIPBlocks.String())
+	}
```
adisos, line 86: is it necessary to differentiate between these cases in the report?

ShiriMoran: I think it is more readable this way. If its practically the same subnet the information is relevant.

pkg/linter/lintOverlappingSubnets.go
```diff
@@ -0,0 +1,115 @@
...
+			return err1
+		}
+		for _, subnet2 := range allSubnets[i+1:] {
+			subnet2IPBlock, err2 := ipblock.FromCidr(subnet2.CIDR())
+			if err2 != nil {
+				return err2
+			}
+			intersectIPBlock := subnet1IPBlock.Intersect(subnet2IPBlock)
+			if !intersectIPBlock.IsEmpty() {
+				if subnetStr(subnet1) > subnetStr(subnet2) {
```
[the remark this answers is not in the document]
ShiriMoran: 1. The comparison is for determinism. 
2. We could, but the intersection result is needed anyways (for the finding) so using the Overlap func instead will only be less efficient, since we will be repeating the same computation twice.

pkg/linter/lintOverlappingSubnets.go
```diff
@@ -0,0 +1,116 @@
...
+		if config.IsMultipleVPCsConfig {
+			continue
+		}
+		allSubnets = append(allSubnets, config.Subnets...)
+	}
+	for i, subnet1 := range allSubnets {
+		subnet1IPBlock, err1 := ipblock.FromCidr(subnet1.CIDR())
+		if err1 != nil {
+			return err1
+		}
```
adisos, line 52: can use `subnet.AddressRange()` instead?

ShiriMoran: right. done.

pkg/linter/lintOverlappingSubnets.go
```diff
@@ -46,15 +46,10 @@ func (lint *overlappingSubnetsLint) check() error {
 		allSubnets = append(allSubnets, config.Subnets...)
 	}
 	for i, subnet1 := range allSubnets {
-		subnet1IPBlock, err1 := ipblock.FromCidr(subnet1.CIDR())
-		if err1 != nil {
-			return err1
-		}
+		subnet1.AddressRange()
```
adisos, line 49: delete this line

ShiriMoran: oops. deleted

adisos (approved):

PR #703, opened 2024-07-18
Issue 702

adisos:
Fix parser - check for nil fields in TransitConnection objects from config input


pkg/ibmvpc/parser.go
```diff
@@ -951,6 +952,11 @@ func getTgwObjects(c *datamodel.ResourcesContainerModel,
 
 	tgwConnList := slices.Clone(c.TransitConnectionList)
 	for i, tgwConn := range c.TransitConnectionList {
+		if tgwConn.TransitGateway.Crn == nil || tgwConn.TransitGateway.Name == nil || tgwConn.NetworkID == nil {
+			logging.Warnf("skipping TransitConnection %s, missing TransitGateway crn/name/networkID", *tgwConn.Name)
```
zivnevo, line 956: what if `tgwConn.Name` is nil as well?

adisos: updated with additional check

zivnevo (approved):

PR #722, opened 2024-07-23
540 lint non connected tables new

ShiriMoran:


pkg/linter/lintFilterRuleSplitSubnet.go
```diff
@@ -84,7 +82,7 @@ func ruleSplitSubnet(subnet vpcmodel.Subnet, ruleIPBlocks []*ipblock.IPBlock) bo
 //////////////////////////////////////////////////////////
 
 func (finding *splitRuleSubnet) vpc() []string {
-	return []string{finding.vpcName}
+	return []string{finding.splitSubnets[0].VPC().Name()}
```
haim-kermany, line 85: 1. rename function to vpcName()
2. why does this function return a slice? if there is a reason, rename it to vpcsNames()

pkg/linter/lintFilterRuleSplitSubnet.go
```diff
@@ -117,8 +115,10 @@ func (finding *splitRuleSubnet) toJSON() any {
 	for i, splitSubnet := range finding.splitSubnets {
 		splitSubnetsJSON[i] = subnetJSON{Name: splitSubnet.Name(), CIDR: splitSubnet.CIDR()}
 	}
-	res := splitRuleSubnetJSON{VpcName: finding.vpcName, Rule: vpcmodel.RuleOfFilter{LayerName: rule.LayerName,
-		FilterName: rule.FilterName, RuleIndex: rule.RuleIndex, RuleDesc: rule.RuleDesc},
+	table := vpcmodel.Filter{LayerName: rule.Filter.LayerName,
+		FilterName: rule.Filter.FilterName}
+	res := splitRuleSubnetJSON{VpcName: finding.splitSubnets[0].VPC().Name(), Rule: vpcmodel.RuleOfFilter{Filter: table,
```
haim-kermany, line 120: use vpcName() here

pkg/linter/lintRedundantTables.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"fmt"
+
+	"github.com/np-guard/vpc-network-config-analyzer/pkg/vpcmodel"
+)
+
+const redundantTablesName = "redundant tables"
+const securityGroup = "security group"
+const networkACL = "network ACL"
```
haim-kermany, line 17: these two last constants are already declared in another package.
expose and reuse.
also fix FilterKindName() to use these constants

pkg/linter/lintRedundantTables.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, config := range lint.configs {
+		if config.IsMultipleVPCsConfig {
+			continue // no use in executing lint on dummy vpcs
+		}
+		for _, layer := range vpcmodel.FilterLayers {
+			filterLayer := config.GetFilterTrafficResourceOfKind(layer)
+			layerName := securityGroup
+			if layer == vpcmodel.NaclLayer {
+				layerName = networkACL
+			}
```
haim-kermany, line 53: use FilterKindName() here

pkg/linter/lintRedundantTables.go
```diff
@@ -0,0 +1,87 @@
...
+			_, _ = filterLayer, layerName
```
haim-kermany, line 54: what is the concept of this line?

haim-kermany (changes requested):

pkg/linter/lintFilterRuleSplitSubnet.go
ShiriMoran: It returns a slice since in multi-vpc configs there could be more than one relevant vpc. 
The interface was agreed upon with @zivnevo; however, reading your comment and revisiting this function:
1. Wouldn't it indeed make more sense to return the VPC and not its name?
2. Why is this function - targeting the VPC - specifically needed in the interface? and if it is, why not also have similar function for other resources relevant to the issue, say Subnets() and Nodes()? @zivnevo

pkg/linter/lintRedundantTables.go
ShiriMoran: concept of "forget to delete it" :-)
to begin with it was to avoid compilation error in the process

pkg/linter/lintFilterRuleSplitSubnet.go
ShiriMoran: After discussing with @zivnevo we agreed to modify the interface so that `vpc()` returns `vpc` instead of name

ShiriMoran: done

pkg/linter/lintFilterRuleSplitSubnet.go
ShiriMoran: done (using `vpc()`)

pkg/linter/lintRedundantTables.go
ShiriMoran: deleted

pkg/linter/lintRedundantTables.go
ShiriMoran: 1. Right. Using `FilterKindName()` no need for the constants here
2. `FilterKindName()` can not use these constants since they are defined in `ibmvpc/commonvpc`
3. Perhaps we should modify the code in `ibmvpc/commonvpc` to use `FilterKindName()` - but not in this PR @adisos do you want to open a followup issue to handle this?

pkg/linter/lintRedundantTables.go
ShiriMoran: done see comment above

pkg/linter/lintRedundantTables.go
adisos: yes, please open a separate issue for this.

ShiriMoran: #724

haim-kermany (approved):

pkg/vpcmodel/abstractVPC.go
```diff
@@ -249,20 +249,26 @@ type RulesInTable struct {
...
 func NewRuleOfFilter(layerName, filterName, desc string, filterIndex, ruleIndex int, ipBlocks []*ipblock.IPBlock) *RuleOfFilter {
-	return &RuleOfFilter{LayerName: layerName, FilterIndex: filterIndex, FilterName: filterName,
-		RuleIndex: ruleIndex, RuleDesc: desc,
-		IPBlocks: ipBlocks}
+	table := Filter{LayerName: layerName, FilterIndex: filterIndex, FilterName: filterName}
+	return &RuleOfFilter{Filter: table, RuleIndex: ruleIndex, RuleDesc: desc, IPBlocks: ipBlocks}
 }
 
+// FiltersToAttached map from each filter to the resources attached to it; e.g. from NACL to subnets
+type FiltersToAttached map[Filter][]VPCResourceIntf
```
adisos, line 270: rename `FiltersToAttached` to `FiltersToAttachedResources` or `FiltersAttachedResources`

pkg/vpcmodel/abstractVPC.go
```diff
@@ -249,20 +249,26 @@ type RulesInTable struct {
...
 
 func NewRuleOfFilter(layerName, filterName, desc string, filterIndex, ruleIndex int, ipBlocks []*ipblock.IPBlock) *RuleOfFilter {
-	return &RuleOfFilter{LayerName: layerName, FilterIndex: filterIndex, FilterName: filterName,
-		RuleIndex: ruleIndex, RuleDesc: desc,
-		IPBlocks: ipBlocks}
+	table := Filter{LayerName: layerName, FilterIndex: filterIndex, FilterName: filterName}
+	return &RuleOfFilter{Filter: table, RuleIndex: ruleIndex, RuleDesc: desc, IPBlocks: ipBlocks}
 }
 
+// FiltersToAttached map from each filter to the resources attached to it; e.g. from NACL to subnets
```
adisos, line 269:
```suggestion
// FiltersToAttached is a map from each filter to the resources attached to it; e.g. from NACL to subnets
```

adisos (commented):
looks good, just added a small comment

adisos (approved):

PR #733, opened 2024-07-28
create internet gateway resource

haim-kermany:
added internet gateway as a resource
creating it during parsing flow.
next step will be to support public subnets


pkg/awsvpc/parser.go
```diff
@@ -291,6 +300,49 @@ func (rc *AWSresourcesContainer) getSGconfig(
...
+		}
+		vpc := res.GetVPC(vpcUID).(*commonvpc.VPC)
+		subnets := vpc.Subnets()
+		// todo - use only the public subnet:
+		if len(subnets) == 0 {
+			logging.Warnf("skipping internet gateway %s - it does not have any attached subnet\n", igwName)
+			continue
+		}
+		routerIgw := newIGW(igwName, igwID, subnets, vpc)
+		// TODO - where to put this resource?
```
adisos, line 325: what does this TODO mean?

pkg/awsvpc/parser.go
```diff
@@ -291,6 +300,49 @@ func (rc *AWSresourcesContainer) getSGconfig(
...
+			logging.Warnf("skipping internet gateway %s - it has %d vpcs attached\n", igwName, len(igw.Attachments))
+			continue
+		}
+		vpcUID := *igw.Attachments[0].VpcId
+		if skipByVPC[vpcUID] {
+			continue
+		}
+		vpc := res.GetVPC(vpcUID).(*commonvpc.VPC)
+		subnets := vpc.Subnets()
+		// todo - use only the public subnet:
```
adisos, line 319: is this planned to a follow up PR?

adisos (commented):
looks good, added few small questions

pkg/awsvpc/parser.go
haim-kermany: removed the todo comment 
(it was not clear if the itg was per region or per vpc, )

pkg/awsvpc/parser.go
haim-kermany: yes

adisos (approved):

PR #749, opened 2024-08-04
Drawio for aws

haim-kermany:
fixing aws set of images
image


haim-kermany:
> what was the problem? what is the fix?

it is not a fix, its implementation for aws.
1. the squares are at the right color and have the right images.
2. the position of the aws squares images are different that the ibm squares

pkg/drawio/styles.go
```diff
@@ -189,20 +208,31 @@ func (stl *templateStyles) setCanTypeHaveAMiniIcon(nodes []TreeNodeInterface) {
...
 // ////////////////////////////////////////////////////////////////////////////////////////
+func (stl *templateStyles) representingType(tn TreeNodeInterface) reflect.Type {
+	if stl.provider != common.AWS || reflect.TypeOf(tn).Elem() != reflect.TypeOf(SubnetTreeNode{}) {
+		return reflect.TypeOf(tn).Elem()
+	}
+	if tn.(*SubnetTreeNode).IsPublic() {
+		return reflect.TypeOf(publicSubnetTreeNode{})
+	}
+	return reflect.TypeOf(privateSubnetTreeNode{})
+}
```
zivnevo, line 219: This looks somewhat ugly.
Would it make more sense to export `publicSubnetTreeNode` and `privateSubnetTreeNode` as specializations of `SubnetTreeNode` instead of having the public flag?

zivnevo (approved):
Feel free to ignore the below comment

PR #755, opened 2024-08-06
Public subnets for igw

haim-kermany:
only public subnets, and thier nodes,  can be the src/dst for igw connections.
the next step is to do some code re-org for routers structs.

still left to do:
1. the existing tests already check igw, however. after merging nacl to main, I will add some complex test that karen suggested.
2. igw egress only option is for ipv6 only, so it will be ignored


pkg/vpcmodel/abstractVPC.go
```diff
@@ -209,6 +209,8 @@ type VPC interface {
 type Subnet interface {
 	NodeSet
 	CIDR() string
+	IsPublic() bool
+	SetIsPublic(bool)
```
adisos, line 213: is this relevant as part of the abstract interface?

haim-kermany: It is an aws concept, However, I thought that it will be needed at explainability (and drawio).
Now, I am not sure that these functions are the right interface for explainability.
I added support for drawio, and see and eventually they are not needed for drawio.
so I deleted the functions from the interface, for now.
see my fix

PR #764, opened 2024-08-08
Lint testing enable disable

ShiriMoran:


pkg/ibmvpc/analysis_output_test.go
```diff
@@ -68,6 +68,8 @@ type vpcGeneralTest struct {
 	EDstMinPort    int64
 	EDstMaxPort    int64
 	detailExplain  bool
+	enable         []string
+	disable        []string
```
zivnevo, line 72: It would make much more sense if each subcommand will extend a basic Testcase struct with its own required fields.
Currently this struct is a mess, as it serves report, diff, explain and lint

haim-kermany (approved):

pkg/ibmvpc/analysis_output_test.go
ShiriMoran: Lets have it in a separate PR.
Some of the struct and functionality is common, and doing this separation properly will require some software engineering work  which is not related to this PR.

zivnevo: OK. Please open an issue

zivnevo (approved):

pkg/ibmvpc/analysis_output_test.go
ShiriMoran: https://github.com/np-guard/vpc-network-config-analyzer/issues/765
Please connect to a milestone

Issue #756, opened 2024-08-06
check aws explainabiliy


haim-kermany:
after nacl is supported, explainability should work.
additional support is needed for public/private subnets

- [x] support public subnets
- [x] detailed explain of public subnets
- [x] adding test infrastructure for aws explainability
- [x] check correctness of examples  + adding interesting tests (html is a great platform for it)

ShiriMoran:
add examples:
1. Private subnet s.t. the path in/to external is being blocked in the subnet
2. Private subnet blocking  both in egress and ingress
3. Public subnets enabling connection to the internet

Issue #655, opened 2024-06-27
explainability - cases in which only partial info is provided


ShiriMoran:
There are two cases in explainability in which only partial information is provided.
These are cases of no connection in which only the reason of "no-connection" is provided.
These are the cases
1. Missing router to external address; the reason only partial info is provided here is that without the router it is not known whether NACL tables are relevant, and so we do not know what are the relevant information
2. Endpoints in two different VPCs with no transit gateway between them. The issue here is development oriented: our code is `VPCConfig` oriented, and due to the lack of the transit gateway there is no `VPCConfig` that contains the relevant endpoints.

ShiriMoran:
Item 1 - the reason is not valid #751 - thus once we fix #762 this can be fixed as well

PR #775, opened 2024-08-12
aws public subnet explainability

haim-kermany:
implementation of the first bullet of:
https://github.com/np-guard/vpc-network-config-analyzer/issues/756


pkg/vpcmodel/explainabilityPrint.go
```diff
@@ -206,6 +213,12 @@ func (g *groupedConnLine) explainPerCaseStr(c *VPCConfig, src, dst EndpointElem,
 	case crossVpcRouterRequired(src, dst) && crossVpcRouter != nil && crossVpcConnection.IsEmpty():
 		return fmt.Sprintf("%vAll connections will be blocked since transit gateway denies route from source to destination"+tripleNLVars,
 			noConnection, headerPlusPath, details)
+	case isAtPrivateSubnet(dst) && src.IsExternal():
```
ShiriMoran, line 216: The cases in which we provide a single line explaining why there is no connection should be minimized see #655. In this specific case we can provide full explainability and integrate the private subnet in the path and details. 

We could defer the explanation part of this PR to latter PRs - I can take responsibility to this part.

ShiriMoran (changes requested):
1. Please add tests
2. See my comment regarding explanation

haim-kermany:
> 1. Please add tests
> 
>     2. See my comment regarding explanation

tests will be added in a different PR
please follow issue https://github.com/np-guard/vpc-network-config-analyzer/issues/756

pkg/vpcmodel/explainabilityPrint.go
haim-kermany: OK, will be resolved in another PR, by our explainability expert.
added a bullet to the relevant issue https://github.com/np-guard/vpc-network-config-analyzer/issues/756

adisos (approved):

pkg/vpcmodel/explainabilityPrint.go
ShiriMoran: Perhaps better remove it all together

ShiriMoran (approved):

PR #788, opened 2024-08-13
Pip tcp response

haim-kermany:
in case of load balancer deny connection, we still needs to collect the connectivity per layer.
so I moved the lb rule check to the end.


pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -169,6 +164,11 @@ func (c *VPCConfig) getAllowedConnsPerDirection(isIngress bool, capturedNode Nod
 			}
 			allLayersRes[peerNode] = routerConnRes
 		}
+		loadBalancerRule := c.getLoadBalancerRule(src, dst)
+		if !isIngress && loadBalancerRule != nil && loadBalancerRule.Deny() {
+			allLayersRes[peerNode] = NoConns()
```
adisos, line 169: should there be another layer for `LB`? 
here you assign the result to `allLayersRes`, but it is not reflected in any of the `perLayerRes`.

zivnevo (approved):

pkg/vpcmodel/nodesConnectivity.go
adisos: opened a separate issue #793

ShiriMoran (approved):

ShiriMoran (approved):
I did not verify the tests

Issue #522, opened 2024-04-11
support virtual network interfaces 


adisos:
https://cloud.ibm.com/docs/vpc?topic=vpc-vni-about

PR #804, opened 2024-08-15
Add vni example

haim-kermany:
connectivity wise, for now, VNI are just like NI.
however, at the parser,  they are extracted from vsi.attachments, and have different icons.
handling first bullet of issue https://github.com/np-guard/vpc-network-config-analyzer/issues/522


pkg/ibmvpc/parser.go
```diff
@@ -380,27 +381,58 @@ func (rc *IBMresourcesContainer) getInstancesConfig(
 		vpcConfig := res.Config(vpcUID)
 		vpcConfig.NodeSets = append(vpcConfig.NodeSets, vsiNode)
 		vpcConfig.UIDToResource[vsiNode.ResourceUID] = vsiNode
+		// a VSI can not have both VNIs and NIs.
+		// first we check for VNIs.
+		// the VNIs are listed at the instance.NetworkAttachments.
+		// using that list, we extract the VNIs from the rc.VirtualNIList, which holds all the VNIs of all VSIs
+		for j := range instance.NetworkAttachments {
+			vniObj := vnisObj[*instance.NetworkAttachments[j].ID]
```
zivnevo, line 389: Need to check we do not get a `nil` here (might happen when running with input files made by an older version of the collector)

pkg/ibmvpc/parser.go
```diff
@@ -380,27 +381,58 @@ func (rc *IBMresourcesContainer) getInstancesConfig(
...
+			continue
+		}
 		for j := range instance.NetworkInterfaces {
 			netintf := instance.NetworkInterfaces[j]
 			// netintf has no CRN, thus using its ID for ResourceUID
-			intfNode, err := commonvpc.NewNetworkInterface(*netintf.Name, *netintf.ID,
-				*instance.Zone.Name, *netintf.PrimaryIP.Address, *instance.Name, len(instance.NetworkInterfaces), vpc)
+			err := createNetworkInterface(*netintf.Name, *netintf.ID,
+				*instance.Zone.Name, *netintf.PrimaryIP.Address, *instance.Name, false, vsiNode, len(instance.NetworkInterfaces),
+				*netintf.Subnet.CRN, subnetIDToNetIntf, vpc, vpcConfig)
```
zivnevo, line 408: better pass just `instance` instead of `*instance.Zone.Name`, `*instance.Name` and `len(instance.NetworkInterfaces)`

zivnevo (changes requested):

pkg/ibmvpc/parser.go
haim-kermany: we can range over nil on golang

zivnevo: I meant what if `vnisObj[*instance.NetworkAttachments[j].ID]` returns `nil`...

pkg/ibmvpc/parser.go
zivnevo: I think it's ok to always use `len(instance.NetworkInterfaces)` because of the guaranteed backwards compatibility.

haim-kermany: fixed. (there are two calls, one with len(instance.NetworkInterfaces) and the other with len(instance.NetworkAttachments))

haim-kermany: fixed

pkg/ibmvpc/parser.go
haim-kermany: in such case, we have a vins that we can not parse.
we can not use the nis instead, since their ids are not at the filters rules
fix to return error

zivnevo (approved):

pkg/ibmvpc/parser.go
```diff
@@ -380,27 +381,62 @@ func (rc *IBMresourcesContainer) getInstancesConfig(
 		vpcConfig := res.Config(vpcUID)
 		vpcConfig.NodeSets = append(vpcConfig.NodeSets, vsiNode)
 		vpcConfig.UIDToResource[vsiNode.ResourceUID] = vsiNode
+		// a VSI can not have both VNIs and NIs.
```
ShiriMoran, line 384: This is guaranteed by the API?

ShiriMoran (commented):
LGTM, two small questions

pkg/ibmvpc/parser.go
haim-kermany: yes

pkg/vpcmodel/multi_vpc_config.go
```diff
@@ -56,7 +57,7 @@ func (c *MultipleVPCConfigs) SetConfigsToCompare(toCompare map[string]*VPCConfig
 	c.toCompareConfigs = toCompare
 }
 func (c *MultipleVPCConfigs) CloudName() string {
-	return string(c.provider) + " Cloud"
+	return strings.ToUpper(string(c.provider)) + " Cloud"
```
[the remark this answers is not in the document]
haim-kermany: not relevant, but not too big for seperate PR

Issue #803, opened 2024-08-15
extend explainability to support aws internal/external subnet


ShiriMoran:
insert "internal subnet" into full explainability path when it blocks connection to or from external internet

PR #828, opened 2024-08-21
Private subnet connectivity and explainability

haim-kermany:
the info about the influence of the privacy of the subnet on the connectivity is capture in a rule (like the lb rule solution)
this rule is used for connectivity and for explainability
the rule is created by the subnet.
it is implemented as part of the vpccommon, since we do not have a different implementation for aws subnet


pkg/commonvpc/vpc.go
```diff
@@ -179,15 +189,68 @@ func (s *Subnet) AddressRange() *ipblock.IPBlock {
...
+	src, dst  vpcmodel.Node
+	isIngress bool
+}
+
+func newPrivateSubnetRule(subnet vpcmodel.Subnet, src, dst vpcmodel.Node, isIngress bool) vpcmodel.PrivateSubnetRule {
+	return &privateSubnetRule{subnet, src, dst, isIngress}
+}
+
+func (psr *privateSubnetRule) Deny(isIngress bool) bool {
+	return isIngress == psr.isIngress && psr.subnet.IsPrivate()
```
ShiriMoran, line 219: This is a bit confusing: here you also have to verify that `src/dst` is external
Specifically, for `ingress`, `src` has to be external and for `egress`, `dst` has to be external
I understand you are not doing it here since the rule will be nil if src and dst are both internal.

Please document this. Something in the spirit:
// Note that privateSubnetRule is defined (non nil) only when relevant: public/private subnet supported by the platform and src/dst is external

pkg/commonvpc/vpc.go
```diff
@@ -179,15 +189,68 @@ func (s *Subnet) AddressRange() *ipblock.IPBlock {
...
+}
+
+func (psr *privateSubnetRule) String() string {
```
ShiriMoran, line 222: I'll finalize the wording and format here in my PR which will link the rule to explainabity

pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -78,6 +78,29 @@ func (c *VPCConfig) getLoadBalancerRule(src, dst Node) LoadBalancerRule {
 	return nil
 }
 
+func (c *VPCConfig) getPrivateSubnetRule(src, dst Node) PrivateSubnetRule {
+	switch {
+	case dst.IsInternal():
```
ShiriMoran, line 83: Wouldn't it be more efficient and also more readable to check here External instead of Internal?
(If both are internal instead of returning nil directly you call `GetPrivateSubnetRule` which return nil in this case)

pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -78,6 +78,29 @@ func (c *VPCConfig) getLoadBalancerRule(src, dst Node) LoadBalancerRule {
...
+		return dst.(InternalNodeIntf).Subnet().GetPrivateSubnetRule(src, dst)
+	case src.IsInternal():
+		return src.(InternalNodeIntf).Subnet().GetPrivateSubnetRule(src, dst)
+	}
+	return nil
+}
+
+// getOtherRulesConn() return the connectivity of all rules that are not part of the filters and routers.
+func (c *VPCConfig) getOtherRulesConn(src, dst Node, isIngress bool) *connection.Set {
```
ShiriMoran, line 92: I don't like the term `other` here.
Perhaps` NonFilterNonRouter`?

pkg/vpcmodel/nodesConnectivity.go
```diff
@@ -164,11 +187,8 @@ func (c *VPCConfig) getAllowedConnsPerDirection(isIngress bool, capturedNode Nod
 			}
 			allLayersRes[peerNode] = routerConnRes
 		}
-		loadBalancerRule := c.getLoadBalancerRule(src, dst)
-		if loadBalancerRule != nil && loadBalancerRule.Deny(isIngress) {
-			allLayersRes[peerNode] = NoConns()
-			continue
-		}
+		otherRulesConn := c.getOtherRulesConn(src, dst, isIngress)
```
ShiriMoran, line 190: same comment regarding `other`

ShiriMoran (commented):
Looks good. Few comments

pkg/commonvpc/vpc.go
```diff
@@ -179,15 +189,68 @@ func (s *Subnet) AddressRange() *ipblock.IPBlock {
 	return s.IPblock
 }
 func (s *Subnet) IsPrivate() bool {
-	return s.isPrivate
+	return s.subnetExpose == privateExpose
 }
 func (s *Subnet) SetIsPrivate(isPrivate bool) {
```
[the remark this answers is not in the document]
haim-kermany: this is not true, IsPrivate() is called for all providers. (for example, for subnet connectivity)
I added a comment

pkg/commonvpc/vpc.go
haim-kermany: tnx

pkg/vpcmodel/nodesConnectivity.go
haim-kermany: the creating a rule is done inside the subnet, the subnet is the struct that "know" the logic of creating the rule.
this code is in the node-connectivity calculation. it should not know anything about the logic of creating the rule.
this switch is only for capturing the subnet.
in ideal world (which has no golang)   GetPrivateSubnetRule() would be a static function of the subnet struct. and this switch was not necessary.

pkg/vpcmodel/nodesConnectivity.go
haim-kermany: fixed

pkg/vpcmodel/nodesConnectivity.go
haim-kermany: here I replaces "other" with "more" to keep it short

pkg/commonvpc/vpc.go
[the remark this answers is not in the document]
ShiriMoran: The function I am referring to here is `SetIsPrivate`

pkg/commonvpc/vpc.go
ShiriMoran: Please add the following comment before this func:
// Note that the func is called only when relevant (platform supporting private subnet and connection to/from internet)

pkg/commonvpc/vpc.go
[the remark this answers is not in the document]
haim-kermany: fixed

pkg/commonvpc/vpc.go
haim-kermany: fixed

ShiriMoran (approved):

pkg/commonvpc/vpc.go
```diff
@@ -179,15 +190,74 @@ func (s *Subnet) AddressRange() *ipblock.IPBlock {
...
+}
+
+func (psr *privateSubnetRule) String() string {
+	switch {
+	case psr.Deny(false):
+		return fmt.Sprintf("%s will not accept connection from %s, since subnet %s is private\n",
+			psr.dst.Name(), psr.src.Name(), psr.subnet.Name())
+	case psr.Deny(true):
+		return fmt.Sprintf("%s will not connect to %s, since subnet %s is private\n",
+			psr.src.Name(), psr.dst.Name(), psr.subnet.Name())
```
zivnevo, line 236:
```suggestion
	case psr.Deny(false), psr.Deny(true):
		return fmt.Sprintf("all traffic from %s to %s is denied, since subnet %s is private\n",
			psr.src.Name(), psr.dst.Name(), psr.subnet.Name())
```

pkg/commonvpc/vpc.go
```diff
@@ -179,15 +190,74 @@ func (s *Subnet) AddressRange() *ipblock.IPBlock {
...
+	case !psr.isIngress:
+		return fmt.Sprintf("%s can accept connection from %s, since subnet %s is public\n",
+			psr.dst.Name(), psr.src.Name(), psr.subnet.Name())
+	case psr.isIngress:
+		return fmt.Sprintf("%s can connect to %s, since subnet %s is public\n",
+			psr.src.Name(), psr.dst.Name(), psr.subnet.Name())
```
zivnevo, line 242:
```suggestion
	default:
		return fmt.Sprintf("traffic from %s to %s is allowed, since subnet %s is public\n",
			psr.src.Name(), psr.dst.Name(), psr.subnet.Name())
```

pkg/commonvpc/vpc.go
```diff
@@ -153,14 +153,25 @@ func (v *VPC) Subnets() []*Subnet {
...
 	vpcmodel.VPCResource
-	VPCnodes []vpcmodel.Node  `json:"-"`
-	Cidr     string           `json:"-"`
-	IPblock  *ipblock.IPBlock `json:"-"`
-	// isPrivate is relevant only for aws, the user set for each subnet if it public (i.e. - has access to the internet)
-	isPrivate bool `json:"-"`
+	VPCnodes     []vpcmodel.Node  `json:"-"`
+	Cidr         string           `json:"-"`
+	IPblock      *ipblock.IPBlock `json:"-"`
+	subnetExpose SubnetExpose     `json:"-"`
```
zivnevo, line 174:
```suggestion
	subnetExpose SubnetExpose
```
No need to prevent json marshaling if a data member is not exported (starts with a small letter)

pkg/commonvpc/vpc.go
haim-kermany: this function is to be edit by shiri, for explainability purposes.
I separated all the cases so she could easily change the text.
she will need a separate line for each case.

pkg/commonvpc/vpc.go
haim-kermany: same here

pkg/commonvpc/vpc.go
haim-kermany: fixed

pkg/commonvpc/vpc.go
zivnevo: @ShiriMoran , I'm not sure which Issue/PR is going to fix the texts here, but currently they are somewhat clunky, and do not match the terminology we usually use.

zivnevo (approved):

pkg/commonvpc/vpc.go
ShiriMoran: This [issue](#803) which is planned for the coming release.
This text is to be used by explainability, and only by it (at least at this stage). 

However, @haim-kermany lets insert @zivnevo's suggestions. I will revisit this soon when working on #803

pkg/commonvpc/vpc.go
ShiriMoran: same here - lets insert Ziv's suggestions @haim-kermany

ShiriMoran (commented):
@haim-kermany please insert Ziv's suggestions and then merge

PR #838, opened 2024-08-28
support multi-vpc in synthesis output

olasaadi99:


kyorav:
When applied to an example that uses a transit gateway the output contains the string `combined-vpc-local-tg1/` instead of the name of the VPC. Example config object attached (taken from the `transit-gateway-multiple` example)
[config_object.json](https://github.com/user-attachments/files/16798787/config_object.json)

kyorav:
I tested on several examples and it looks great

pkg/vpcmodel/output.go
```diff
@@ -301,14 +301,51 @@ func (of *serialOutputFormatter) AggregateVPCsOutput(outputList []*SingleAnalysi
...
+		}
+		if conn.Dst.Type == spec.ResourceTypeExternal {
+			if val, ok := externalsMap[conn.Dst.Name]; ok {
+				conn.Dst.Name = val
+			} else {
+				name := externalString + strconv.Itoa(len(externalsMap))
+				externalsMap[conn.Dst.Name] = name
+				conn.Dst.Name = name
+			}
+		}
```
zivnevo, line 343:
```suggestion
		if conn.Src.Type == spec.ResourceTypeExternal {
			conn.Src.Name = getNewExternalName(conn.Src.Name, externalsMap)
		} else if conn.Dst.Type == spec.ResourceTypeExternal {
			conn.Dst.Name = getNewExternalName(conn.Dst.Name, externalsMap)
		}
```
Avoid code duplication

pkg/vpcmodel/output.go
```diff
@@ -301,14 +301,51 @@ func (of *serialOutputFormatter) AggregateVPCsOutput(outputList []*SingleAnalysi
...
 
+func renameExternals(requiredConnections []spec.SpecRequiredConnectionsElem,
+	externalsMap map[string]string) []spec.SpecRequiredConnectionsElem {
+	connLines := []spec.SpecRequiredConnectionsElem{}
+	for _, conn := range requiredConnections {
+		if conn.Src.Type == spec.ResourceTypeExternal {
+			if val, ok := externalsMap[conn.Src.Name]; ok {
+				conn.Src.Name = val
+			} else {
+				name := externalString + strconv.Itoa(len(externalsMap))
```
zivnevo, line 330: `fmt.Sprintf()` might be better here

zivnevo (changes requested):

zivnevo (approved):

PR #848, opened 2024-09-03
support bidirectional synthesis

olasaadi99:


pkg/vpcmodel/synthesisOutput.go
```diff
@@ -59,25 +59,54 @@ func getSynthesisSpec(groupedLines []*groupedConnLine) *spec.Spec {
...
 	s.Externals = externals
 	s.RequiredConnections = connLines
 	return &s
 }
 
+func initBidirectionalMap(groupedLines []*groupedConnLine, bidirectionalMap map[string]bool) {
+	for _, groupedLine := range groupedLines {
+		_, ok := bidirectionalMap[groupedLine.Src.SynthesisResourceName()+
+			groupedLine.Dst.SynthesisResourceName()+
+			groupedLine.CommonProperties.Conn.allConn.String()]
```
zivnevo, line 96: It's dangerous to create a map key by concatenating strings without separators.
`{src: "ab", dst: "cde"}` will have the same key as `{src: "abc", dst: "de"}`

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -59,25 +59,54 @@ func getSynthesisSpec(groupedLines []*groupedConnLine) *spec.Spec {
...
+			Bidirectional:    bidirectional})
 	}
 	s.Externals = externals
 	s.RequiredConnections = connLines
 	return &s
 }
 
+func initBidirectionalMap(groupedLines []*groupedConnLine, bidirectionalMap map[string]bool) {
+	for _, groupedLine := range groupedLines {
+		_, ok := bidirectionalMap[groupedLine.Src.SynthesisResourceName()+
```
zivnevo, line 94: Suggesting a function to create the map key from a groupedLine (with a parameter whether or not to flip src and dst)

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -59,25 +59,54 @@ func getSynthesisSpec(groupedLines []*groupedConnLine) *spec.Spec {
...
+		if ok {
+			bidirectionalMap[groupedLine.Src.SynthesisResourceName()+
+				groupedLine.Dst.SynthesisResourceName()+
+				groupedLine.CommonProperties.Conn.allConn.String()] = true
+		} else {
+			// insert to map opposite direction
+			bidirectionalMap[groupedLine.Dst.SynthesisResourceName()+
+				groupedLine.Src.SynthesisResourceName()+
+				groupedLine.CommonProperties.Conn.allConn.String()] = false
+		}
```
zivnevo, line 106: The logic here confuses me.
It seems like we insert entries for (possibly) non-existing connections.
It turns out all-right because later we search for this flipped connection, but is there a way to make things more straightforward?

zivnevo (changes requested):

pkg/vpcmodel/synthesisOutput.go
olasaadi99: I can insert the correct direction in this else.. but this way it is more easier to get just one of the bidirectional connections in the map

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -61,23 +62,63 @@ func getSynthesisSpec(groupedLines []*groupedConnLine) *spec.Spec {
...
+			AllowedProtocols: sortProtocolList(spec.ProtocolList(connection.ToJSON(groupedLine.CommonProperties.Conn.allConn))),
+			Bidirectional:    bidirectional})
 	}
 	s.Externals = externals
 	s.RequiredConnections = connLines
 	return &s
 }
 
+func getBidirectionalMapKeyByNames(firstName, secName, conn string) string {
+	return fmt.Sprintf("%s-%s-%s", firstName, secName, conn)
```
zivnevo, line 92: Better use a character that cannot be part of an endpoint name, such as `_` or `~`.
See the name field [here](https://cloud.ibm.com/apidocs/vpc/latest#create-vpc) for the list of legal characters.

pkg/vpcmodel/synthesisOutput.go
```diff
@@ -61,23 +62,63 @@ func getSynthesisSpec(groupedLines []*groupedConnLine) *spec.Spec {
...
+func makeBidirectionalMap(groupedLines []*groupedConnLine) map[string]bool {
+	bidirectionalMap := make(map[string]bool)
+	for _, groupedLine := range groupedLines {
+		_, ok := bidirectionalMap[getBidirectionalMapKeyByConnLine(groupedLine, false)]
+		if ok {
+			bidirectionalMap[getBidirectionalMapKeyByConnLine(groupedLine, false)] = true
+		} else {
+			// insert to map opposite direction
+			bidirectionalMap[getBidirectionalMapKeyByConnLine(groupedLine, true)] = false
+		}
```
zivnevo, line 117:
```suggestion
		_, ok := bidirectionalMap[getBidirectionalMapKeyByConnLine(groupedLine, true)]
		if ok { // reverse direction is already in the map - just mark it bidirectional
			bidirectionalMap[getBidirectionalMapKeyByConnLine(groupedLine, true)] = true
		} else {
			bidirectionalMap[getBidirectionalMapKeyByConnLine(groupedLine, false)] = false
		}
```
This logic is more intuitive for me (you'll also have to flip the check later).

zivnevo (changes requested):

zivnevo (approved):

PR #856, opened 2024-09-05
767 sg nacl fields print order

ShiriMoran:


cmd/analyzer/expected_out/acl_testing3_detailed_explain.txt
```diff
@@ -15,25 +15,25 @@ Details:
 Path is enabled; The relevant rules are:
 	Egress:
 		security group sg1-ky allows connection with the following allow rules
-			direction: outbound, id: id:152, remote: 0.0.0.0/0, local: 0.0.0.0/0, conns: protocol: all
+			id: id:152, direction: outbound, local: 0.0.0.0/0, remote: 0.0.0.0/0, conns: protocol: all
```
zivnevo, line 18:
```suggestion
			id: id:152, direction: outbound, local: 0.0.0.0/0, remote: 0.0.0.0/0, protocol: all
```

cmd/analyzer/expected_out/acl_testing3_detailed_explain.txt
```diff
@@ -15,25 +15,25 @@ Details:
...
 		network ACL acl1-ky allows connection with the following allow and deny rules
-			direction: outbound, name: acl1-out-1, priority: 1, action: deny, source: 10.240.10.0/24 , destination: 10.240.20.0/24, conn: protocol: icmp
-			direction: outbound, name: acl1-out-3, priority: 3, action: allow, source: 10.240.10.0/24 , destination: 10.240.20.0/24, conn: all
+			name: acl1-out-1, priority: 1, action: deny, direction: outbound, source: 10.240.10.0/24, destination: 10.240.20.0/24, conn: protocol: icmp
+			name: acl1-out-3, priority: 3, action: allow, direction: outbound, source: 10.240.10.0/24, destination: 10.240.20.0/24, conn: all
```
zivnevo, line 21:
```suggestion
			name: acl1-out-1, priority: 1, action: deny, direction: outbound, source: 10.240.10.0/24, destination: 10.240.20.0/24, protocol: icmp
			name: acl1-out-3, priority: 3, action: allow, direction: outbound, source: 10.240.10.0/24, destination: 10.240.20.0/24, protocol: all
```

zivnevo (changes requested):

cmd/analyzer/expected_out/acl_testing3_detailed_explain.txt
ShiriMoran: Lets fix this properly in connections. Set by adding "protocol" before "all" and then indeed the "conns" is redundant
I'll do it in a followup PR.
For now I'm removing the "conns", which will look weird when all protocols are in the connection.
This is to avoid an extra PR.

ShiriMoran: done

cmd/analyzer/expected_out/acl_testing3_detailed_explain.txt
ShiriMoran: done

pkg/ibmvpc/nacl_analysis.go
```diff
@@ -74,15 +75,16 @@ func (na *IBMNACLAnalyzer) GetNACLRule(index int) (ruleStr string, ruleRes *comm
...
+		portsStr = fmt.Sprintf(", srcPorts: %s, dstPorts: %s", srcPorts, dstPorts)
 		direction = *ruleObj.Direction
 		src = *ruleObj.Source
 		dst = *ruleObj.Destination
 		action = *ruleObj.Action
 	case *vpc1.NetworkACLRuleItemNetworkACLRuleProtocolIcmp:
 		name = *ruleObj.Name
 		conns = commonvpc.GetICMPconn(ruleObj.Type, ruleObj.Code)
-		connStr = fmt.Sprintf("protocol: %s", *ruleObj.Protocol)
+		protocol = *ruleObj.Protocol
```
zivnevo, line 87: shouldn't you also set `portsStr` to show code and type (if relevant)?

pkg/awsvpc/nacl_analysis.go
```diff
@@ -72,7 +72,7 @@ func (na *AWSNACLAnalyzer) GetNACLRule(index int) (ruleStr string, ruleRes *comm
 			minPort,
 			maxPort,
 		)
-		connStr = fmt.Sprintf("protocol: %s, dstPorts: %d-%d", protocol, minPort, maxPort)
+		portsStr = fmt.Sprintf(", dstPorts: %d-%d", minPort, maxPort)
 	case protocolICMP:
```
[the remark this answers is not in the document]
ShiriMoran: 1. Indeed, there is no such example; moreover,  I did not find an example that enables only icmp in the inputs file.  Perhaps such an example should be added; does not seem related to this PR.
2. Looking at the code - indeed it seems that the extra ICMP parms are ignored in the returned string; they were ignored from the beginning, nothing new to this PR. 
3. Moreover, looking at the code I do not understand why the printing function of conns was not used in the first place for the returned `ruleStr`. 

Not sure whether to fix it here (as I think it should be) or open a new issue and ask Ola what she had in mind.

pkg/ibmvpc/nacl_analysis.go
ShiriMoran: Guess we should. see comment above.
I would change the code so that it uses the printing function of conn; not sure why it was not done this way in the first place; perhaps I am missing something.

pkg/awsvpc/nacl_analysis.go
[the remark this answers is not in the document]
ShiriMoran: fixed; At the moment we do not have an example of an icmp rule in aws

We do have an example in IBM (acl5) that is not yet used in explain tests; I'll add a test

pkg/ibmvpc/nacl_analysis.go
ShiriMoran: fixed, see comment above

pkg/ibmvpc/nacl_analysis.go
```diff
@@ -74,23 +75,31 @@ func (na *IBMNACLAnalyzer) GetNACLRule(index int) (ruleStr string, ruleRes *comm
...
 		direction = *ruleObj.Direction
 		src = *ruleObj.Source
 		dst = *ruleObj.Destination
 		action = *ruleObj.Action
+		if ruleObj.Type != nil && *ruleObj.Type != -1 {
+			portsStr = fmt.Sprintf("icmp type: %d", *ruleObj.Type)
+		}
+		if ruleObj.Code != nil && *ruleObj.Code != -1 {
+			portsStr += fmt.Sprintf("icmp code: %d", *ruleObj.Code)
+		}
```
zivnevo, line 97:
```suggestion
		if ruleObj.Type != nil {
			portsStr = fmt.Sprintf(", type: %d", *ruleObj.Type)
		}
		if ruleObj.Code != nil {
			portsStr += fmt.Sprintf(", code: %d", *ruleObj.Code)
		}
```
The -1 value is an AWS-only thing

pkg/awsvpc/nacl_analysis.go
```diff
@@ -72,20 +72,26 @@ func (na *AWSNACLAnalyzer) GetNACLRule(index int) (ruleStr string, ruleRes *comm
...
 
 		if err != nil {
 			return "", nil, false, err
 		}
+		if ruleObj.IcmpTypeCode.Type != nil && *ruleObj.IcmpTypeCode.Type != -1 {
+			portsStr = fmt.Sprintf("icmp type: %d", *ruleObj.IcmpTypeCode.Type)
+		}
+		if ruleObj.IcmpTypeCode.Code != nil && *ruleObj.IcmpTypeCode.Code != -1 {
+			portsStr += fmt.Sprintf("icmp code: %d", *ruleObj.IcmpTypeCode.Code)
+		}
```
zivnevo, line 88:
```suggestion
		if ruleObj.IcmpTypeCode.Type != nil && *ruleObj.IcmpTypeCode.Type != -1 {
			portsStr = fmt.Sprintf(", type: %d", *ruleObj.IcmpTypeCode.Type)
		}
		if ruleObj.IcmpTypeCode.Code != nil && *ruleObj.IcmpTypeCode.Code != -1 {
			portsStr += fmt.Sprintf(", code: %d", *ruleObj.IcmpTypeCode.Code)
		}
```

zivnevo (changes requested):

pkg/awsvpc/nacl_analysis.go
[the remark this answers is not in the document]
ShiriMoran: acl5 does not have any instances

Need to add examples of `NACL` rules with reference to `ICMP`  type and code in both `IBM` and `AWS`

zivnevo (approved):

PR #865, opened 2024-09-10
refactor Name() and ExtendedName(c) methods

olasaadi99:


pkg/vpcmodel/abstractVPC.go
```diff
@@ -21,6 +21,7 @@ const (
 type VPCResourceIntf interface {
 	UID() string
 	Name() string
+	NameForAnalyzerOut() string
```
adisos, line 24: still confusing/unclear functionality for `Name` / `NameForAnalyzerOut` / `ExtendedName` / `ExtendedPrefix`.

- Does the interface need to have all these methods?
- what is the use case for each method?
- which methods are overriden for certain types? is this required? (cannot be implemented explicitly instead of changing expected behavior of certain method?) 
- is there a method that always returns the original resource name from the config object?

adisos (changes requested):
added a comment with some questions.

pkg/vpcmodel/abstractVPC.go
olasaadi99: > * is there a method that always returns the original resource name from the config object?

Name()
consider renaming to resourceName() + doc

> Does the interface need to have all these methods?

we need Name() and NameForAnalyzerOut , ExtendedName and  ExtendedPrefix will check if we can merge it with NameForAnalyzerOut

pkg/ibmvpc/vpc.go
```diff
@@ -66,20 +66,21 @@ type PrivateIP struct {
 	block *ipblock.IPBlock
 }
 
-func (pip *PrivateIP) Name() string {
+func (pip *PrivateIP) NameForAnalyzerOut(c *vpcmodel.VPCConfig) string {
+	prefix := ""
+	if c != nil && c.IsMultipleVPCsConfig {
+		prefix = pip.VPC().Name() + vpcmodel.Deliminator
+	}
```
adisos, line 73: this code seems to be duplicated across this file.. can be moved to some other internal function that gets config and VPC name as inputs?

pkg/vpcmodel/abstractVPC.go
```diff
@@ -67,8 +59,12 @@ func (n *VPCResource) SynthesisKind() spec.ResourceType {
 	return ""
 }
 
-func (n *VPCResource) ExtendedName(c *VPCConfig) string {
-	return n.ExtendedPrefix(c) + n.Name()
+func (n *VPCResource) NameForAnalyzerOut(c *VPCConfig) string {
+	prefix := ""
+	if c != nil && c.IsMultipleVPCsConfig {
```
adisos, line 64: maybe the config type should have a method that encapsulates this concept of required prefix?

adisos (commented):
looks good!
few more comments

adisos:
one more comment: can you revisit where `NameForAnalyzerOut ` is called with `nil` and where not?
when is it required? when it can be assumed it is not needed?
are there places where the config can actually be retrieved but currently not used in the call? and if so, why?

olasaadi99:
> one more comment: can you revisit where `NameForAnalyzerOut ` is called with `nil` and where not? when is it required? when it can be assumed it is not needed? are there places where the config can actually be retrieved but currently not used in the call? and if so, why?

I assumed that it is not needed where we originally called the previous func without the config... and passed the config where we originaly called ExtendedName (with the config).

adisos:
> > one more comment: can you revisit where `NameForAnalyzerOut ` is called with `nil` and where not? when is it required? when it can be assumed it is not needed? are there places where the config can actually be retrieved but currently not used in the call? and if so, why?
> 
> I assumed that it is not needed where we originally called the previous func without the config... and passed the config where we originaly called ExtendedName (with the config).

makes sense, but if you can revisit as suggested above this may help to better understand these aspects, update documentation as to where this is expected to be used, etc.. (maybe also find if it makes sense to add the config in other places, to be less error-prone).

pkg/ibmvpc/vpc.go
```diff
@@ -39,8 +39,8 @@ type ReservedIP struct {
 
 func (r *ReservedIP) NameForAnalyzerOut(c *vpcmodel.VPCConfig) string {
 	prefix := ""
-	if c != nil && c.IsMultipleVPCsConfig {
-		prefix = r.VPC().Name() + vpcmodel.Deliminator
+	if c != nil {
+		prefix = c.MultipleVPCsConfigPrefix(r.VPC().Name())
```
adisos, line 43: can add another internal function that gets as input arg the config object and returns that prefix, and here always just have `prefix + nameWithBracketsInfo(r.vpe, r.Address())`
(to avoid this duplication across this file as well) .

pkg/vpcmodel/vpcConfig.go
```diff
@@ -44,6 +44,14 @@ type VPCConfig struct {
 	IsMultipleVPCsConfig bool
 }
 
+// MultipleVPCsConfigPrefix returns the passed prefix when config is multi-vpc
+func (c *VPCConfig) MultipleVPCsConfigPrefix(prefix string) string {
```
adisos, line 48: change arg name to vpcName ?

pkg/commonvpc/vpc.go
```diff
@@ -737,3 +733,11 @@ func getTableConnEffect(connQuery, conn *connection.Set) (*connection.Set, vpcmo
 		return conn.Intersect(connQuery), vpcmodel.PartlyAllow
 	}
 }
+
+// MultipleVPCsConfigPrefix returns the passed vpcName when config is multi-vpc
+func MultipleVPCsConfigPrefix(c *vpcmodel.VPCConfig, vpcName string) string {
+	if c != nil && c.IsMultipleVPCsConfig {
```
adisos, line 739:
```suggestion
	if c != nil {
```

pkg/commonvpc/vpc.go
```diff
@@ -737,3 +733,11 @@ func getTableConnEffect(connQuery, conn *connection.Set) (*connection.Set, vpcmo
 		return conn.Intersect(connQuery), vpcmodel.PartlyAllow
 	}
 }
+
+// MultipleVPCsConfigPrefix returns the vpcName of the passed resource when config is multi-vpc
+func MultipleVPCsConfigPrefix(c *vpcmodel.VPCConfig, resource *vpcmodel.VPCResource) string {
+	if c != nil && resource.VPC() != nil {
```
adisos, line 739: when is `resource.VPC()` expected to be `nil`?

adisos (commented):
looks good.
There are more cases with `NameForAnalyzerOut(nil)` is used. Do you have insights about those cases?

pkg/commonvpc/vpc.go
olasaadi99: I got some tests failed before I added this condition, it seems like there are some cases we do not init config file in tests

olasaadi99:
> looks good. There are more cases with `NameForAnalyzerOut(nil)` is used. Do you have insights about those cases?

in these cases the config param is not reachable, or that we need to make a lot of changes

pkg/commonvpc/vpc.go
adisos: only unit tests or end-to-end tests?
this is not about initializing config file, but having the resource a pointer to the VPC it is contained in.

adisos:
> > looks good. There are more cases with `NameForAnalyzerOut(nil)` is used. Do you have insights about those cases?
> 
> in these cases the config param is not reachable, or that we need to make a lot of changes

@ShiriMoran can you review the places where `NameForAnalyzerOut(nil)` is now used in this PR, and whether it is safe to keep these calls, or whether the full name (with vpc prefix for multi-vpc cases) should be used instead?  (such as explainability output)

pkg/vpcmodel/abstractVPC.go
```diff
@@ -67,8 +59,12 @@ func (n *VPCResource) SynthesisKind() spec.ResourceType {
 	return ""
 }
 
-func (n *VPCResource) ExtendedName(c *VPCConfig) string {
-	return n.ExtendedPrefix(c) + n.Name()
+func (n *VPCResource) NameForAnalyzerOut(c *VPCConfig) string {
+	prefix := ""
+	if c != nil && n.VPC() != nil {
```
ShiriMoran, line 64: if `c != nil`, can `n.VPC()` be nil?

ShiriMoran:
> NameForAnalyzerOut(nil)

1. Syntactic and backwards compatibility wise, it is safe. Namely, all the places in which `NameForAnalyzerOut(nil)` are places in which `Name()` was used before
2. There are places in which  `Name()` was used before and now `NameForAnalyzerOut(c)` is called now, which is good
3. This leaves us in some mode of confusion, in which there seems to be no rule, or one that makes sense, as to when we avoid the `VPC` prefix 
4. The above rule is "places where before we used `Name()` in functions that do not have the `VPC config`". I think we should avoid this as much as possible, namely avoid calling `NameForAnalyzerOut` with nil.
5. Thus, we should go over all the cases in which  `NameForAnalyzerOut(nil)` is called and try to replace it with  `NameForAnalyzerOut(c)`; this will require adding vpc config parm to the relevant functions. This can be done in a followup PR @adisos @olasaadi99 (or in this one, what ever you prefer)

ShiriMoran:
> @ShiriMoran can you review the places where `NameForAnalyzerOut(nil)` is now used in this PR, and whether it is safe to keep these calls, or whether the full name (with vpc prefix for multi-vpc cases) should be used instead? (such as explainability output)


1. Syntactic and backwards compatibility wise, it is safe. Namely, all the places in which `NameForAnalyzerOut(nil)` are places in which `Name()` was used before
2. There are places in which  `Name()` was used before and now `NameForAnalyzerOut(c)` is called now, which is good
3. This leaves us in some mode of confusion, in which there seems to be no rule, or one that makes sense, as to when we avoid the `VPC` prefix 
4. The above rule is "places where before we used `Name()` in functions that do not have the `VPC config`". I think we should avoid this as much as possible, namely avoid calling `NameForAnalyzerOut` with nil.
5. Thus, we should go over all the cases in which  `NameForAnalyzerOut(nil)` is called and try to replace it with  `NameForAnalyzerOut(c)`; this will require adding vpc config parm to the relevant functions. This can be done in a followup PR @adisos @olasaadi99 (or in this one, what ever you prefer)

adisos:
> > @ShiriMoran can you review the places where `NameForAnalyzerOut(nil)` is now used in this PR, and whether it is safe to keep these calls, or whether the full name (with vpc prefix for multi-vpc cases) should be used instead? (such as explainability output)
> 
>     1. Syntactic and backwards compatibility wise, it is safe. Namely, all the places in which `NameForAnalyzerOut(nil)` are places in which `Name()` was used before
> 
>     2. There are places in which  `Name()` was used before and `NameForAnalyzerOut(c)` is called now, which is good
> 
>     3. This leaves us in some mode of confusion, in which there seems to be no rule, or one that makes sense for a user, as to when we avoid the relevant `VPC` prefix
> 
>     4. The above rule is "places where before we used `Name()` in functions that do not have the `VPC config`". I think we should avoid this as much as possible, namely avoid calling `NameForAnalyzerOut` with `nil`.
> 
>     5. Thus, we should go over all the cases in which  `NameForAnalyzerOut(nil)` is called and try to replace it with  `NameForAnalyzerOut(c)`; this will require adding `vpc config` parm to the relevant functions. This can be done in a followup PR @adisos @olasaadi99 (or in this one, what ever you prefer)

Thanks @ShiriMoran , sounds good. I think so as well.
@olasaadi99  please check if you prefer a separate PR for this.

olasaadi99:
prefer new pr

adisos (approved):

PR #878, opened 2024-09-17
fix current tests with overlapping address prefixes + Remove function

olasaadi99:
fix current tests with overlapping address prefixes + Remove function validateAddressPrefixesExist()

### DIFF
--- a/pkg/ibmvpc/examples/input/input_multi_regions.json
+++ b/pkg/ibmvpc/examples/input/input_multi_regions.json
@@ -62,6 +62,21 @@
             "resource_type": "vpc",
             "status": "available",
             "region": "us-east",
+            "address_prefixes": [
+                {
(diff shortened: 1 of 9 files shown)


zivnevo:
How did you determine the `address_prefixes` field of each VPC?
Is this based on a new run of the collector, manually from subnet info, something else?

olasaadi99:
> How did you determine the `address_prefixes` field of each VPC? Is this based on a new run of the collector, manually from subnet info, something else?

all of them based on a new run from terrafotm examples except for `tgw_larger_example` based on subnet info (`address_prefixes`  for this examples from terraform examples are with overlapped, and all of the tests assumes that it is not overlapping and use the subnet cidrs as AP)

zivnevo (approved):

PR #883, opened 2024-09-19
Use the new map/slice features in Go 1.23

### LINKED ISSUE #763: revisit genericSet.go when go1.23 is avaliable

zivnevo:
Fixes #763

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/np-guard/vpc-network-config-analyzer
 
-go 1.22.4
+go 1.23.1
 
 require (
 	github.com/IBM/networking-go-sdk v0.49.0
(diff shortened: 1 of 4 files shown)


haim-kermany (commented):
1. consider sunseting the methods MapValues() and MapKeys(). its look like they are  just unnecessary wrappers 
2. use keys() also for AnyMapEntry() (or sunset AnyMapEntry) too )

zivnevo:
> 1. consider sunseting the methods MapValues() and MapKeys(). its look like they are  just unnecessary wrappers
> 
>     2. use keys() also for AnyMapEntry() (or sunset AnyMapEntry) too )

1. Done.
2. I actually like `AnyMapEntry()`, and I think it is useful. I don't see a nicer way to write it with `Keys()` or `Values()`

haim-kermany (approved):

PR #891, opened 2024-09-25
887 integrate lint in flow

ShiriMoran:
First stage of integrating lint #891 - separating the execution to analysis and print


pkg/linter/linterExecute.go
```diff
@@ -84,14 +94,20 @@ func LinterExecute(configs map[string]*vpcmodel.VPCConfig, printAllFindings bool
...
+			return nil, err
 		}
+	}
+	return linters, nil
+}
+
+func (linters Linters) String(printAllFindings bool) (resString string) {
+	strPerLint := []string{}
+	for _, thisLinter := range linters {
+		thisLintStr := ""
```
zivnevo, line 108:
```suggestion
```
Redundant line

pkg/linter/linterExecute.go
```diff
@@ -65,17 +65,27 @@ func computeConnectivity(configs map[string]*vpcmodel.VPCConfig) (map[string]*vp
...
+// integrated in the general flow
 func LinterExecute(configs map[string]*vpcmodel.VPCConfig, printAllFindings bool,
 	enableList, disableList []string) (issueFound bool, resString string, err error) {
-	nodesConn, err := computeConnectivity(configs)
+	linters, err := linterAnalysis(configs, enableList, disableList)
 	if err != nil {
 		return false, "", err
 	}
+	resString = linters.String(printAllFindings)
+	fmt.Println(resString)
```
zivnevo, line 77: `linters.String()` also has a `Println` for the results - so results are printed twice?

pkg/linter/linterExecute.go
```diff
@@ -65,17 +65,27 @@ func computeConnectivity(configs map[string]*vpcmodel.VPCConfig) (map[string]*vp
...
+	return issueFound, resString, nil
```
zivnevo, line 78: Where do you set `issuesFound`?

zivnevo (changes requested):

pkg/linter/linterExecute.go
ShiriMoran: its not, and not really need. removed (the function was supposed to be deleted before PR I did not bother in the first place)

pkg/linter/linterExecute.go
ShiriMoran: I removed the `Println` from `linters.String()`
I still believe  `LinterExecute` will be deleted soon

pkg/linter/linterExecute.go
ShiriMoran: right. removed (and replace = in  := where needed)

zivnevo (approved):

PR #895, opened 2024-09-30
print cidr in graph in case of a single cidr

ShiriMoran:


pkg/vpcmodel/grouping.go
```diff
@@ -643,22 +643,32 @@ func listEndpointElemStrWithConfig(eps []EndpointElem, fn func(ep EndpointElem,
 }
 
 func (g *groupedExternalNodes) String() string {
+	// 1 gets list and a union of all IPBlocks
+	ipbList, unionBlock := g.toIPBlocks()
+	for _, ipBlock := range ipbList {
+		unionBlock = unionBlock.Union(ipBlock)
+	}
```
zivnevo, line 650: Why are we repeating the Union here? We already got it from `toIPBlocks`.

pkg/vpcmodel/drawioGenerator.go
```diff
@@ -106,6 +106,11 @@ func (g *groupedExternalNodes) GenerateDrawioTreeNode(gen *DrawioGenerator) draw
 	name := "Various IP ranges"
 	if all, _ := isEntirePublicInternetRange(*g); all {
 		name = publicInternetNodeName
+	} else {
+		_, ipBlock := g.toIPBlocks()
+		if len(ipBlock.ToCidrList()) == 1 {
```
zivnevo, line 111: How about the case where we have a single IP range that cannot be represented as a single CIDR?
Can we show the IP range instead of "Various IP ranges"?

zivnevo (changes requested):

pkg/vpcmodel/drawioGenerator.go
ShiriMoran: yes, done

pkg/vpcmodel/grouping.go
```diff
@@ -643,22 +643,28 @@ func listEndpointElemStrWithConfig(eps []EndpointElem, fn func(ep EndpointElem,
...
+func (g *groupedExternalNodes) toIPBlocks() *ipblock.IPBlock {
 	// 1. Created a list of IPBlocks
 	cidrList := make([]string, len(*g))
 	for i, n := range *g {
 		cidrList[i] = n.CidrStr
 	}
 	ipbList, _, err := ipStringsToIPblocks(cidrList)
 	if err != nil {
-		return ""
+		return nil
```
zivnevo, line 660: I think it's better to return an empty IPBlock, to avoid nil dereferencing issues

pkg/vpcmodel/grouping.go
```diff
@@ -643,22 +643,28 @@ func listEndpointElemStrWithConfig(eps []EndpointElem, fn func(ep EndpointElem,
 }
 
 func (g *groupedExternalNodes) String() string {
+	// 1 gets list and a union of all IPBlocks
+	unionBlock := g.toIPBlocks()
+	// 2. print a list s.t. each element contains either a single cidr or an ip range
+	return strings.Join(unionBlock.ListToPrint(), commaSeparator)
+}
+
+func (g *groupedExternalNodes) toIPBlocks() *ipblock.IPBlock {
```
zivnevo, line 652:
```suggestion
func (g *groupedExternalNodes) toIPBlock() *ipblock.IPBlock {
```
To be more consistent

zivnevo (changes requested):

pkg/vpcmodel/grouping.go
ShiriMoran: done

pkg/vpcmodel/grouping.go
ShiriMoran: ok, done

zivnevo (approved):

haim-kermany (approved):

PR #932, opened 2024-11-27
removing function AnalyzeNACLRules and fix tests

olasaadi99:


adisos (approved):

pkg/ibmvpc/naclAnalysis_test.go
```diff
@@ -46,112 +51,303 @@ func testSingleNACL(nacl *commonvpc.NACL) {
...
+			fmt.Printf("ipRange.ToIPRanges(): %v\n", ipRange.ToIPRanges())
+			fmt.Printf("conn: %v\n", conn)
+		}
 
-	fmt.Printf("done\n")
+		fmt.Printf("connectivityResult.DenyRules: \n")
+		for ipRange, conn := range connectivityResult.DenyRules {
+			fmt.Printf("ipRange.ToIPRanges(): %v\n", ipRange.ToIPRanges())
+			fmt.Printf("conn: %v\n", conn)
+		}
```
zivnevo, line 342: Do we need all these prints?
We already have huge logs for our tests - it makes it hard to search for failures.
So unless there is a failure, I prefer tests to produce only very little text if at all.

zivnevo (changes requested):

pkg/ibmvpc/naclAnalysis_test.go
olasaadi99: not needed
removed

olasaadi99: in https://github.com/np-guard/vpc-network-config-analyzer/pull/933

